PR #2, opened 2017-07-23
Engines registration

Nadahar:
The PR refactors the way engines/```Player```s are registered.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/2)


src/main/java/net/pms/encoders/PlayerFactory.java
```diff
@@ -442,6 +478,189 @@ public static Player getPlayer(final DLNAResource resource) {
...
+			} catch (IOException e) {
+				LOGGER.warn("\"{}\" failed with error: {}", executable.toString(), e.getMessage());
+				LOGGER.trace("", e);
+				state = String.format(Messages.getString("Engine.Error"), player) + " \n" + e.getMessage();
+				player.setUnavailable(state);
+			}
+			TEST_RECORDS.add(new PlayerTestRecord(executable, player.isAvailable(), state));
+			return true;
+		}
+		// No test has been made for VLC, found no way to get feedback on stdout: https://forum.videolan.org/viewtopic.php?t=73665
```
Sami32: Do you mean something like that ?
https://trac.videolan.org/vlc/ticket/17286

Nadahar: @Sami32 Something like that - I don't remember the exact details now, but I think couldn't find a way to get some simple feedback saying "VLC is installed and working" without it launching the GUI and having to be closed manually by the user.

VLC isn't very much used as an engine, so I don't consider it a big deal though.

src/main/java/net/pms/newgui/ViewLevel.java
```diff
@@ -21,166 +21,169 @@
 /**
  * Defines a set of view levels for changing the complexity of the GUI:<br>
  * {@link #Normal}, {@link #Advanced}, {@link #Expert} and {@link #Unknown}.
- * <p/> The <code>ViewLevel</code> class is final and cannot be sub-classed.</p>
+ * <p>
+ * This class is final and cannot be sub-classed.
```
Sami32: "sub-classed" or [subclassed](https://docs.oracle.com/javase/tutorial/java/IandI/final.html) seem to be well spreaded, though the second form is a little more used.
It's just a comment, as i don't think that anything is wrong with your writing.

src/main/java/net/pms/configuration/ExternalProgramInfo.java
```diff
@@ -0,0 +1,382 @@
...
+			lock.readLock().unlock();
+		}
+	}
+
+	/**
+	 * Sets the path for a given {@link ProgramExecutableType} for this
+	 * instance.
+	 *
+	 * @param executableType the {@link ProgramExecutableType} whose path to
+	 *            set.
```
Nadahar: @Sami32 From [what I can understand](http://www.dictionary.com/browse/whose) it's correct. Dropping "whose" would give the sentence no meaning.

src/main/java/net/pms/encoders/FFmpegWebVideo.java
```diff
@@ -69,15 +69,15 @@
...
+		Path executable = Paths.get(_configuration.getFFmpegPath());
 		try {
 			FilePermissions permissions = new FilePermissions(executable);
 			if (permissions.isExecutable()) {
 				protocolsLock.writeLock().lock();
 				try {
 					FFmpegOptions.getSupportedProtocols(protocols, executable.toString());
 					if (protocols.contains("mmsh")) {
-						// Workaround an FFmpeg bug: http://ffmpeg.org/trac/ffmpeg/ticket/998
+						// Work around an FFmpeg bug: http://ffmpeg.org/trac/ffmpeg/ticket/998
```
Nadahar: @Sami32 Argh - I didn't see the "an". "workaround" isn't correct here AFAICU - the reason is that "work" is used as a verb here: Please work around that bug. "workaround" is a noun and thus the sentence wouldn't make any sense.

src/main/java/net/pms/newgui/ViewLevel.java
Nadahar: I'll refactor that class completely, it can be simplified a lot by turning it into an ```enum```. When it comes to splitting it into two words, it's probably Eclipse's dictionary that complained (it does that for a lot of terms I'd say is well accepted as one word). I've started to add them instead of letting Eclipse "decide", but I'm betting that's the reason.

src/main/java/net/pms/encoders/FFmpegWebVideo.java
Sami32: Aaah i wrongly understood it as a noun. So, yes, you're right.

src/main/java/net/pms/configuration/ExternalProgramInfo.java
Sami32: Your link have more clear definition than mine. Yes, that look fine. Now it's seem similar to the "which" word to me.

Nadahar: @Sami32 It's very similar to "of which".

src/main/java/net/pms/newgui/ViewLevel.java
Sami32: i didn't even knew that Eclipse have an integrated dictionary 8-/

Nadahar: Eclipse has almost everything 😉

src/main/java/net/pms/encoders/FFmpegWebVideo.java
Nadahar: Btw, I think both your Java and English skills improve 😉

src/main/java/net/pms/newgui/ViewLevel.java
Sami32: :stuck_out_tongue:

src/main/java/net/pms/encoders/PlayerFactory.java
```diff
@@ -442,6 +478,189 @@ public static Player getPlayer(final DLNAResource resource) {
...
+			} catch (IOException e) {
+				LOGGER.warn("\"{}\" failed with error: {}", executable.toString(), e.getMessage());
+				LOGGER.trace("", e);
+				status = String.format(Messages.getString("Engine.Error"), player) + " \n" + e.getMessage();
+				player.setUnavailable(status);
+			}
+			TEST_RECORDS.add(new PlayerTestRecord(executable, player.isAvailable(), status));
+			return true;
+		}
+		// No test has been made for VLC, found no way to get feedback on stdout: https://forum.videolan.org/viewtopic.php?t=73665
```
Sami32: What about that ?
`vlc -I none --quiet --file-logging --logfile=vlc16.log vlc://quit`
If VLC is working, that will throw:
```logtalk
-- logger module started --
-- logger module stopped --
```

Nadahar: I don't remember what the exact problem was, but remember that this has to work on Linux and OS X as well.

From a quick test it "works" on Windows, but it still opens and closes a new window. Also, I'd need to get the version printed so I could parse it, but that might be possible.

PS! It's not "throw" but "print". "Throw" is used epecially for Exceptions when they are created and "thrown".

Sami32: OK, isee. So use that one:
`vlc -I dummy --dummy-quiet colored.jpg --file-logging --logfile=vlc.txt --image-duration=0 vlc://quit`

I'll try to remember "print", as i didn't liked "throw" but didn't know what to use instead. Thanks :-)

Nadahar: @Sami32 Bon appétit.

While you get the version in the log file that way, it's a hack 😛 I don't think you'll get rid of the window that opens, I think that's the essential problem (that VLC is made in a very strange way that runs inside this new window). Also, having to read the results from a file isn't optimal, it could lead to all kind of permission issues, and isn't very effective. I guess the temp folder could be used to avoid permission issues, but I still think it's way too much trouble to get the version. 

I actually tried to not be a "perfectionist" here and leave it alone because using VLC for transcoding isn't that relevant these days. You don't make it easy though 😇

Sami32: Sadly true, sorry :flushed: 
I'll try to not do it again. So, just forget about my comment.

Nadahar: @Sami32 Just admit that you're a bit like me.. it's a bit hard to "let something go" if you think you can fix it.. 😏 

Your second command still shows opens a window, and it stores the result in a file. But yes, it would be possible to parse the version from that file.

Sami32: Na :stuck_out_tongue: 
`vlc -I dummy -Vdummy --dummy-quiet --file-logging --logmode=text vlc://quit`
Still not perfect, but no external input need and not external output file :stuck_out_tongue_winking_eye:

Nadahar: @Sami32 The problem is that it's nothing to parse at all.

Sami32: LOL Only in the second console window :P
Or into the default _vlc-log.txt_ file created in the VLC folder.

Nadahar: @Sami32 No, that's the problem. To parse the version (without letting VLC write it to a file and then parse that file) we need to get the output in the window/session that UMS "controls" - that's what can be read from starting the process - which is nothing. Very strange behaviour by VLC, "everybody" else does this.

Sami32: This guy used _libvlc.dll_ to get infos without hassle:
https://github.com/caprica/vlcj
https://github.com/caprica/vlcj/blob/a3637da2bf3d552f71a9f3cce9942da03c8aef34/src/main/java/uk/co/caprica/vlcj/binding/LibVlcFactory.java

Nadahar: @Sami32 Yes, he's using JNA. That's obviously possible, but is a more "heavy" way than I intended.

Sami32: I promisse i stop now to bother you with that 8-/
It's also because i don't see anymore way :wink:

Nadahar: @Sami32 Than we've come to the same conclusion 😉 

It IS possible, but not without a lot of extra hazzle, which doesn't make it worth it simply for displaying the version. DMS still does the "basic test" for VLC, that the binary is found and that it has execute permission.

Nadahar: Hrmf.. damn that you got me to look at this again. On Linux it's very simple: ```vlc --version```.

Sami32: After, the properties of vlc excecutable can also be read to obtain the version number but...LOL

Nadahar: @Sami32 I've actually never looked into how to get that metadata from Java.

Nadahar: @Sami32 Lol, JNA as well: https://stackoverflow.com/questions/6918022/get-version-info-for-exe

Sami32: OK, i should admit that i can be a little obsessional compulsive LOL

Sami32: Yes, but `vlc --version` work only with Linux and macOS, Windows get the display of the console windows :disappointed:

Sami32: I saw the fun of verifying the engines status, as the way to automatically enable / disable them in the configuration at the start of UMS.

Nadahar: @Sami32 This PR does that - it doesn't actually change the "enabled" status, as that's a configuration setting, but it introduces an "available" status as well, which is set by the tests. I've then made something I call "active", which is only true if they are both enabled and available, and that's what determines if they are actually used.

In addition to that, the engines can have 3 "types" of executables: Bundled, installed and custom. Bundles only applies to those that are bundles on that platform, installed usually means that it's reachable via ```PATH``` (although for AviSynth and VLC the registry is checked on Windows). Custom means that the used as configured a specific path to the executable.

In the transcoding setup, it will then be possible to choose what "type" to use, but availability information is gathered for all variants during startup so that you can see the version of each, for example when you switch between bundled and installed FFmpeg.

Sami32:
Shouldn't a flag showing that a PR is ready to review be added ?

Note: I added a "Not ready" flag.

Nadahar:
@sami32 I hope you're happy with r9 and that it does all you want :stuck_out_tongue_winking_eye: 

---

Review status: 0 of 74 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2#-:-KqNU58LWlr_uSuWnHVT:b-t0pqrn)*

src/main/java/net/pms/util/Version.java
```diff
@@ -1,251 +1,792 @@
...
+		 * @throws LastErrorException if an error occurs.
+		 */
+		int GetFileVersionInfoSizeW(String fileName, IntByReference ignore) throws LastErrorException;
+
+		/**
+		 * Retrieves version information for the specified file.
+		 *
+		 * @param fileName the name of the file. If a full path is not
+		 *            specified, the function uses the search sequence specified
+		 *            by by the LoadLibrary function.
```
Sami32: Mistyping: `by by`
> by the LoadLibrary function.

Sami32:
You make me look as a dictator :P
I have just suggested that it could be possible, because you read that it wasn't or didn't found.

That's said, that look great :+1: at least for what i can understand. I'll have to make a PR on my fork for future build test, as for now i'm stuck.

Nadahar:
@Sami32 I just wrote that I hadn't found a way to do it "the usual way", by reading the output from the command line (on Windows). But, I discovered that the VLC version is used for checking that certain features are there, so it is actually needed. The existing implementation of these tests are flawed, as the version is only gotten from the Windows registry and will only be correct on the currently "installed" version in Windows - and fail for Linux and OS X.

It ended up being quite a lot of work though, but I did some other generalizations/improvements as well (for example by making a generic version number parser that can be used to hold and compare the version for any program).

I don't understand why you can't build this though, have you cloned the repository? What is the error/problem?

Sami32:
I saw code related to what you are talking but i do understand it fully.

JNA 4.4.0. Never mind, i just made a revert PR on my clone.

BTW, i just saw that you have white trailing issue in your PRs. I don't understand, as i thought that you've fixed it already, except if you reverted back or that it doesn't work on old PRs amendments ?

Nadahar:
@Sami32 There has been so many JNA issues with your computer that I really don't remember. I did compile a ```jnidispatch.dll``` on my XP VM that I though should work without issues. Did you ever try to use that instead of the one that comes with JNA?

If you have to fork every commit you are to test, it's quite a job. You could also just edit the POM.xml after checking out without commiting it, instead of doing to forking. Still a hazzle though.

About the white trailing issue, the check only applies when I make (new or amended) commits. I've been doing a lot of rebasing a fixing to make this code work with base code that is 1 year newer (I'm still not completely finished, although it's getting closer), so I'd think the possibility for whitespace issues is quite high. That said, CheckStyle reacts on it. Unfortunately I can only use CheckStyle on classes I write from scratch (or refactor heavily), as there is so many warnings in "traditional UMS code" that most of the code is marked with a warning.

About ```Version``` it's not as simple as comparing two values. Your first need to figure out what to compare, and that's what ```Version``` does. It parses the individual parts of a version number and then compares given certain rules, for example that missing number compare as zero (```4.3.0.0 == 4.3``` and ```9.7 > 9.6.0.2 > 9.6```). If you'd just compare version numbers as strings, they would be compared alphabetically, which is useless for version numbers.

Sami32:
I make a modified master, and after i don't need to care anymore, as i will use _update.jar_, as i already do with UMS since then. It's relatively straithforward. The PR is only used with major master changes.

> I did compile a jnidispatch.dll on my XP VM that I though should work without issues. Did you ever try to use that instead of the one that comes with JNA?

Hum...i'll need to try :flushed:

Nadahar:
@Sami32 Using ```update.jar``` that way will lead to bugs if any of the dependencies are changed in the PR. In this case I don't think they are though, so it should work.

Please note that there's a initialization bug in this PR that I haven't fixed yet, that will lead to "A fatal error during startup" or something similar being logged. It also makes the configuration file not work. It's just because things aren't started in the correct order, but I have to untangle a lot of stuff to figure out how to start everything in the correct order.

Sami32:
I remember two others thing that i totally "forgotten":
- I think that come from UMS change, as i didn't make a new build since long time, i use _update.jar_ since the JNA upgrade. After installation it keep asking for a JRE 7.0+, i already have it installed.
- i've already forgotten :P

Nadahar:
@Sami32 I don't quite understand what you mean, but using ```update.jar``` is risky unless the "original" is updated to have the exact same dependencies.

[jnidispatch.dll.txt](https://github.com/DigitalMediaServer/DigitalMediaServer/files/1188609/jnidispatch.dll.txt)

That's the file you need. The way to tell JNA to use that file instead of that shippws with JNA is to set the java property ```jna.boot.library.path``` to point to the file. It can be done in multiple ways, but the simplest is probably to include it in the command line:
```
javaw -Djna.boot.library.path="<pathto>\jnidispatch.dll" -jar dms.jar
```
That means that you can't start DMS using DMS.exe, but you can easily make a ```.bat``` file that makes this call for you each time you want to start DMS. It should be easier that dealing with forks and ```update.jar``` all the time. Remember to fill in the *actual path* to the file instead of ```<pathto>``` 😉

Sami32:
Before playing with that, i'll need to modify the install code, as it seem it doesn't detect correctly the Java version installed, at least in my case.

Nadahar:
@Sami32 What do you mean by "install code"?

Sami32:
capture_07312017_215352

Nadahar:
@Sami32 But why do you run the installer? All you have to do is to build it with maven and then run the command I showed you. ```dms.jar``` can be found in the ```target``` folder after ```mvn package```.

Sami32:
I know, but i'm just trying the standard users way, and that confirm what i read on the forum since the automatic version download PR was done.

Nadahar:
@Sami32 Automatic version download PR? Are you referring to an UMS bug/PR?

Sami32:
https://github.com/Sami32/DigitalMediaServer/commit/fe017ed48d3830ed5982ffa93f21fb15df8a37a5
Yes, users have complained that they need to download Java again at UMS install.

Nadahar:
@Sami32 I don't understand how that can be related to this bug, I've removed ```AutoUpdater``` from DMS.

Nadahar:
@Sami32 It seems like the version comparison in the installer fails somehow. I don't know what has changed, maybe it caused by Java 7 not being the current Java version?

I don't understand the ```nsh``` code, it looks almost like assembler to me. I don't know how it determines what ```javaw.exe``` to check the version of, but it might be the one currently on the path.

The check is this:
```nsh
Function CheckJREVersion
	; R1 holds the current JRE version
	Push $R1
	Push $R2

	; Get the file version of javaw.exe
	${GetFileVersion} $R0 $R1

	ClearErrors

	; Check if JRE7 is installed
	${VersionCompare} ${JRE7_VERSION} $R1 $R2
	StrCmp $R2 "1" 0 CheckDone

	SetErrors

	CheckDone:
		Pop $R1
		Pop $R2
FunctionEnd
```

Which calls this (I bet you think my ```Version``` is easier to understand 😛)

```nsh
!macro VersionCompare_
	!verbose push
	!verbose ${_WORDFUNC_VERBOSE}

	Exch $1
	Exch
	Exch $0
	Exch
	Push $2
	Push $3
	Push $4
	Push $5
	Push $6
	Push $7

	WordFunc_VersionCompare_begin:
	StrCpy $2 -1
	IntOp $2 $2 + 1
	StrCpy $3 $0 1 $2
	StrCmp $3 '' +2
	StrCmp $3 '.' 0 -3
	StrCpy $4 $0 $2
	IntOp $2 $2 + 1
	StrCpy $0 $0 '' $2

	StrCpy $2 -1
	IntOp $2 $2 + 1
	StrCpy $3 $1 1 $2
	StrCmp $3 '' +2
	StrCmp $3 '.' 0 -3
	StrCpy $5 $1 $2
	IntOp $2 $2 + 1
	StrCpy $1 $1 '' $2

	StrCmp $4$5 '' WordFunc_VersionCompare_equal

	StrCpy $6 -1
	IntOp $6 $6 + 1
	StrCpy $3 $4 1 $6
	StrCmp $3 '0' -2
	StrCmp $3 '' 0 +2
	StrCpy $4 0

	StrCpy $7 -1
	IntOp $7 $7 + 1
	StrCpy $3 $5 1 $7
	StrCmp $3 '0' -2
	StrCmp $3 '' 0 +2
	StrCpy $5 0

	StrCmp $4 0 0 +2
	StrCmp $5 0 WordFunc_VersionCompare_begin WordFunc_VersionCompare_newer2
	StrCmp $5 0 WordFunc_VersionCompare_newer1
	IntCmp $6 $7 0 WordFunc_VersionCompare_newer1 WordFunc_VersionCompare_newer2

	StrCpy $4 '1$4'
	StrCpy $5 '1$5'
	IntCmp $4 $5 WordFunc_VersionCompare_begin WordFunc_VersionCompare_newer2 WordFunc_VersionCompare_newer1

	WordFunc_VersionCompare_equal:
	StrCpy $0 0
	goto WordFunc_VersionCompare_end
	WordFunc_VersionCompare_newer1:
	StrCpy $0 1
	goto WordFunc_VersionCompare_end
	WordFunc_VersionCompare_newer2:
	StrCpy $0 2

	WordFunc_VersionCompare_end:
	Pop $7
	Pop $6
	Pop $5
	Pop $4
	Pop $3
	Pop $2
	Pop $1
	Exch $0

	!verbose pop
!macroend
```

Sami32:
It has to do with Java version not correctly detected using the registry key, at least on Windows XP using JDK 7. I'm still searching on that.

Nadahar:
@Sami32 It's not actially using the registry for the version, it's only using the registry to find the path to the ```CurrentVersion``` Java executable. It's then using the file version that of ```javaw.exe```. I have no idea how to debug nsh.

Sami32:
Arrrg, the latest update that i did few day ago only installed the JDK version...that's why ;-)

Nadahar:
@Sami32 Ok, that's strange. I've never seen a JDK that doesn't also install JRE. Anyway, I think I found the code that has been copied for downloading Java: http://nsis.sourceforge.net/Java_Launcher_with_automatic_JRE_installation

As you can see, this is quite old.

Sami32:
OK, that was it :wink: 
Nice looking bythe way :+1: 
The main icon is not my cup of tea, but i don't any idea or anything better to offer.
I really love your other icons, sincerly, they are cuet.

It seem that i understood your code incorrectly, as i thought that double clicking on video / audio engines will select/unselect them, and that's not the case.

Nadahar:
@Sami32 Just to be sure, did you use "my" ```jnidispatch.dll```?

It's nice that you like the buttons, I spent a lot of time making the logo as well, sorry that it isn't your taste. It isn't that easy to make logos, but I actually like it.

You're supposed to be able to doubleclick the transcoding engines to enable/disable them. Doesn't that work? It works for me. Can you enable/disable them with the red/green button?

I'm also interested in if it has correctly registered your "engines" and versions. If you hold the mouse over one of them, a desciption with an error message or the version should appear.

Sami32:
Nothing to worry about, i didn't say that i disliked it, and everyone have its own taste.
I never tried, and i'm very interested in making icon. If my father was still in life i could have enjoyed creating it...

Yes, the buttons work as expected for enabling / disabling the video / audio engines, but not the double clicking from my WinXP at least.

Sorry, no version popup for me, but if its only my XP it doen't really matter.
No error are displayed in the trace log, and i remember having seen the VLC version in the log.

Nadahar:
@Sami32 Do you see a text (tooltip) when you hold the mouse over one of the transcoding engines at all?

Sami32:
Sorry, your messages didn't appeared to me until that i refreshed the tab now.
As i can see...i did searched to see where you have that it was old LOL but yes it have some years ;-)
Happily i get suspicious when i tested in Git the Java version and got a registry error, as i was going to search in the wrong NSIS direction.
I would a say like a mix of C and assembler.

Sami32:
I think that "Engines are in descending order..." should be better displayed as a popup while crossing the entry over the engines windows ?

Unusual warnings are throwed in the log at starting time concerning the renderers:
`WARN  2017-08-01 00:08:08.265 [AWT-EventQueue-0] Renderer Yamaha RX-V671 found twice, ignoring repeated entry`

Sami32:
I finally get suspiscious about JNA so i tried your DLL, but got no changement.

Nadahar:
@Sami32 I'm still confused 😉 

Won't it run when you use my ```jnadispatch.dll```? I'm talking about running ```dms.jar``` directly, without ```update.jar```. Or, did you mean that using my ```jnadispatch.dll``` didn't make any difference for showing the versions? If that's what you mean, it wouldn't. JNA is only used to get the VLC version, not the other engines.

About the suggestion for the tooltip, I'm not sure I understand. Do you see tooltips when hovering over the engine names or not? If not, do you see tooltips at all, when hoveding over other fields or buttong?

I don't think it's possible, or at least not very practical, to have a tooltip for both the whole "engines list" and for each separate engine. Is that what you meant?

As for the GUI, I haven't done all the changes I'm planning with the GUI on engine configuration. I'm still working on getting the other parts right, which is why I am interested in knowing if the versions are parsed correctly.

Nadahar:
@Sami32 I need some ```.dvr-ms``` sample files. Do you know of some, I've had some trouble finding it?

Sami32:
- I meant that using your `jnadispatch.dll` didn't make any difference for showing the versions.
- NO tooltips when hovering over the engine names.
- Yes, the others tooltips are correctly displayed (with green background)
- I meant that this text below look crappy to me and not really lisible; IMO. And tooltips should be better.
capture_08012017_022316

Sami32:
Here is your sapmle:
http://streams.videolan.org/issues/3629/Football_DVRMS_stuttering.dvr-ms

Nadahar:
@Sami32 Ok, that cleared up some of it. It still isn't clear to me if:
* You still use fork + ```update.jar```
* Reinstalling Java made it run using JNA 4.4.0 without change
* It runs when using ```jnadispatch.dll``` when built and run via ```javaw.exe ...``` without any ```update.jar``` or other "tricks".

Sami32:
> You still use fork + update.jar

No, just the master with JNA 4.3.0
> Reinstalling Java made it run using JNA 4.4.0 without change

No, only run. I still use JNA 4.3.0.
> It runs when using jnadispatch.dll when built and run via javaw.exe ... without any update.jar or other "tricks".

Yes, with or without your DLL, its run with the master, as long as i downgrade to JNA 4.3.0.

Nadahar:
@Sami32 My ```jnadispatch.dll``` is built for JNA 4.4.0. It shouldn't even work with JNA 4.3.0, which makes me wonder if it's even loaded.

Sami32:
Aaahh i forgotten that point.
How could verify that it is loaded ?
`start javaw -Xmx768M -Djava.net.preferIPv4Stack=true -Djna.boot.library.path="C:\Program Files\Digital Media Server\jnidispatch.dll" -Dfile.encoding=UTF-8 -classpath update.jar;dms.jar net.pms.PMS`

Nadahar:
@Sami32 I don't know, it would have to be some kind of JNA debugging that I don't know enough about. You could try to drop  ```-Djava.net.preferIPv4Stack=true``` so that there's only less ```-D``` parameter (I don't know if that matters), and use ```/``` instead of ```\``` in the path, but that's just "blind guesses" - your command line does look correct to me. 

I remember one thing that might interfere with it though, there's a setting ```jna.nosys``` that is set to ```true``` in the code during startup - I don't know exactly how those parameters work, so maybe that setting disables whatever is set in ```jna.boot.library.path```.

Sami32:
`jna.nosys` right, i remember that he told us about it. I'll check tomorrow.

Nadahar:
@Sami32 It seems like ```jna.debug_load=true``` should provide some more information of what JNA is doing. However, this won't go into DMS' log file but to the console, so you'll need to start DMS using ```java.exe``` instead of ```javaw.exe``` (and preferrable without ```start```) to see the output.

Nadahar:
@Sami32 [This JavaDoc](https://java-native-access.github.io/jna/4.3.0/javadoc/com/sun/jna/Native.html) seems to confirm my suspicion that ```jna.nosys``` disables ```jna.boot.library.path```:


> 
> 
> ## Library Loading
> 
> When JNA classes are loaded, the native shared library (jnidispatch) is loaded as well. An attempt is made to load it from the any paths defined in jna.boot.library.path (if defined), then the system library path using System.loadLibrary(java.lang.String), unless jna.nosys=true. If not found, the appropriate library will be extracted from the class path (into a temporary directory if found within a jar file) and loaded from there, **unless jna.noclasspath=true**. If your system has additional security constraints regarding execution or load of files (SELinux, for example), you should probably install the native library in an accessible location and configure your system accordingly, rather than relying on JNA to extract the library from its own jar file.

Sami32:
Damned !!! I did a mistake in my command line, it seem that you've missed it as well. Hehe, it was too late :wink: 
`-Djna.boot.library.path="C:\Program Files\Digital Media Server\jnidispatch.dll"`
I shouldn't have added `\jnidispatch.dl`. Now it work fine ;-)

**Note**: I added 4 more DVR-MS sample in an edit below the first one, that should be suffisant ;-)

Nadahar:
@Sami32 Nice, then maybe you should look at #7 - it could be a permanent solution.

Nadahar:
@Sami32 Thanks for the additional samples, but I've already taken your word and removed ```FFmpegDVRMSRemux```. I'm not so worried about breaking stuff here, no release is planned in the near future and getting an updated FFmpeg to work is something that must be done at some stage.

Sami32:
I agree, as it seem already fixed, and that's an very rare and obsolete format, from Microsoft support at least.
We should remember to update FFmpeg and fontconfig.
The cache of fontconfig seem to be the root of the issues affecting some users not having subtitles.

What a busy bee. It very kind to you, but i can make a build without bothering you, as i anyway need to use a special FFmpeg build.
That's also why i think that the engine "valid" detection is useful, as i saw some users using only MEncoder, because FFmpeg and VLC engines was failing on their computer but they didn't know why (only their log give me the explanation that i wasn't supported by their computer or badly installed): and since they didn't complained about that, it's a good thing that it will displayed as they could wonder and ask us, IMO.

Nadahar:
@Sami32 There are several advantages to this as I see it, being able to see if an engine isn't working is one of them. This approach is also much better to "find" engines that aren't bundled, and to let users choose if they want to use the bundled engine or not (GUI not implemented yet).

I've looked at the fontconfig before, and even though I don't look forward to touch it again, our implementation looks less than ideal for me and I can imagine that it can create problems some times. I can't see a reason why it shouldn't be possible to get a recent FFmpeg to work.

About the XP build, it's not *just* for you. If that PR works, we can merge it and it will help other XP users as well.

Sami32:
That worked for me at least, i just updated FFmpeg and fontconfig.
I only saw Linux users having that kind of subtitles issues, but the others didn't attach any log so...

Sorry, i forget to post the log with the error on the way to be fixed that you was talking about:
```logtalk
$ java -Xmx768M -Djava.net.preferIPv4Stack=true -Dfile.encoding=UTF-8 -Djna.debug_load.jna=true -Djna.boot.library.path="C:\Program Files\Digital Media Server" -jar "C:\Program Files\Digital Media Server\dms.jar"
Looking in C:\Program Files\Digital Media Server\jnidispatch.dll
Trying C:\Program Files\Digital Media Server\jnidispatch.dll
Found jnidispatch at C:\Program Files\Digital Media Server\jnidispatch.dll
INFO  14:58:15.156 [System Information Logger] System information:
  JVM: Java HotSpot(TM) Client VM 1.7.0_151 (32-bit) by Oracle Corporation
  OS: Microsoft Windows XP SP3 build 2600 32-bit
  CPU: AMD Athlon(TM) XP 1700+ with 1 core
  Physical Memory: 3.0 GiB
  Free Memory: 2.1 GiB
  Maximum JVM Memory: 742.4 MiB

INFO  14:58:15.609 [main] Starting Digital Media Server 0.0.1-SNAPSHOT
INFO  14:58:15.609 [main] Based on PS3 Media Server by shagrath and Universal Media Server
INFO  14:58:15.609 [main] http://www.digitalmediaserver.org
INFO  14:58:15.609 [main]
INFO  14:58:15.609 [main] Build: af137f63c (2017-07-30)
INFO  14:58:15.609 [main] Java: Java HotSpot(TM) Client VM 1.7.0_151 (32-bit) by Oracle Corporation
INFO  14:58:15.609 [main] OS: Windows XP 32-bit 5.1
INFO  14:58:15.609 [main] Maximum JVM Memory: 742.4 MiB
INFO  14:58:15.625 [main] Language: English (United States)
INFO  14:58:15.625 [main] Encoding: UTF-8
INFO  14:58:15.625 [main]
INFO  14:58:15.625 [main] Working directory: F:\DigitalMediaServer
INFO  14:58:15.671 [main] Temporary directory: C:\DOCUME~1\Philippe\LOCALS~1\Temp\DMS
INFO  14:58:15.671 [main] Logging configuration file: F:\DigitalMediaServer\src\main\external-resources\logback.xml
INFO  14:58:15.671 [main] Logfile: C:\Documents and Settings\All Users\DigitalMediaServer\debug.log
INFO  14:58:15.671 [main]
INFO  14:58:15.671 [main] Profile directory: C:\Documents and Settings\All Users\DigitalMediaServer
INFO  14:58:15.703 [main] Profile directory permissions: drwx
INFO  14:58:15.703 [main] Profile configuration file: C:\Documents and Settings\All Users\DigitalMediaServer\DMS.conf
INFO  14:58:15.703 [main] Profile configuration file permissions: -rwx
INFO  14:58:15.703 [main] Profile name: MAISON
INFO  14:58:15.703 [main]
INFO  14:58:15.703 [main] Web configuration file: C:\Documents and Settings\All Users\DigitalMediaServer\WEB.conf
INFO  14:58:15.703 [main] Web configuration file permissions: -rwx
INFO  14:58:15.703 [main]
INFO  14:58:15.781 [main] Using database located at: "C:\Documents and Settings\All Users\DigitalMediaServer\database"
INFO  14:58:20.796 [main] Checking shared folder: F:\!MEDIA_TEST_FILES\AUDIO
INFO  14:58:20.796 [main] Checking shared folder: F:\!MEDIA_TEST_FILES\UMS Video Testing Suite (20150121)
**ERROR 14:58:23.031 [main] A serious error occurred during Digital Media Server initialization: Unable to establish loopback connection**

```

Nadahar:
@Sami32 I'm not too worried about the "serious error occured" error as I know what's causing it. That said, I haven't seen it complain about the loopback connection before, but it could still be due to the same error. The fundamental problem is that UMS is initialized before the configuration is, which leads to all kind of errors.

Nadahar:
@Sami32 
> I guess that one of these issues explain why no popups version are displayed ?

(I'm answering it in the this PR where it belongs) No, it is the fact that the currently bundled version of FFmpeg doesn't work on XP, and I didn't really like the way it was handled. It ended up as an "unknown error", so I have refined the error parsing so it makes a bit more sense. 

I'm done with that now, but I just found a new problem. I implemented the parsing of Windows' "file metadata``` to get the version information stored there, but it turns out that it doesn't work when the file is on the ```PATH```. The file path you send to the Windows API must either be absolute or relative from the current folder, it won't find executables that is found somewhere on the ```PATH```. So, I'm currently making a way to search the path for a given file from Java, so that I can find the actual location of the executable.

When it comes to the popups, I never understood the problem. You didn't really answer if you:
* Don't see tooltips at all, for any components
* Don't see tooltips in the engines list only
* See tooltips but without any version information

Until I understand the problem, I can't really try to figure out what's causing it.

Sami32:
So now you understand why i use my FFmpeg build, but knowing that, you saw that it give me no popups engines version anyway.

I've already answered you.
> Don't see tooltips at all, for any components

**i see others popups**
> Don't see tooltips in the engines list only

**YES.**
> See tooltips but without any version information

**No see engines tooltips**

Nadahar:
@Sami32 Because of our extreme ability to misunderstand each other, not everything you answer is understood as an answer by my 😉 

I didn't know what the problem was, but now that I know that the tooltips are missing altogheter I can only scratch my head - the version parsing has nothing to do with that, a tooltip should be displayed even if the version isn't parsed, it just won't include the version.

It must be something very fundamental, as the tooltips should "just work". I've done nothing special with them, but I suspect that it could be related to why the doubleclicks doesn't work for you. I'll have to test it on my XP VM and see if I can figure anything out.

Nadahar:
@Sami32 I just tested the latest version of this branch on my XP VM, and the tooltips work just fine with versions and everything. The same goes for doubleclicking. Could you try again with the current version of the branch?

Sami32:
> @Sami32 Because of our extreme ability to misunderstand each other, not everything you answer is understood as an answer by my :wink:
I couldn't have said better :rofl: 

OK, but i will need to solve my Git issue first, as it seem that my remote setting messed ?

Sami32:
My remotes was fine, my tracking was not :flushed: 

Yes, you're right, it is fixed now.
So, what was causing it ?

Sami32:
Btw, now that told about FFmpeg version and WinXP, i remember that it a similar issue for "old" macOS.

Nadahar:
@Sami32 I have no idea what was the issue, it was working for me all the time. I guess you have done some "tricks" (with update.jar or whatever) that made it not work.

When it comes to old OS X, it's a mess. Apple doesn't care about backwards compatibility and has always done what Microsoft has adopted in recent years: Make life as hard as possible for those that doesn't "keep up" and buy the newest eqipment all the time. I think I drew the line at 10.6 when I made the OS X/macOS JNA code. They have changes so much that you'd really have to write several sets of code for many things if you want to keep things working with older versions.

The "good" thing is that later versions of OS X and macOS doesn't cost money to upgrade. The bad thing is as always that the newer versions doesn't support older hardware, so they get your money in any case.

Just to show how little respect they have for standardization, they changed the name of their "10" OS between 10.11 and 10.12. The versions prior to 10.12 is called OS X while 10.12 is called macOS. It makes everything so complicated, I'm tempted to just call it all for "crap" - or maybe "crap10"?

Sami32:
I've asked one year ago, that advertising should be done on the download section of the forum for XP and old Mac OS users, with links to a working version for them (they already freely accessible on the net), but as i never get answer i began to discovered that "Universal" was just a vain word without real will...
At least advertising doesn't cost anything, and help users to understand why not accusing the program for itself and fix it, IMO...

Nadahar:
@Sami32 I completely agree with you. If there is ever a web page for DMS, this should be done.

SubJunk:
> that advertising should be done on the download section of the forum

There is no download section of the forum and I don't know what you mean by "advertising". 

> i began to discovered that "Universal" was just a vain word without any real will...

I don't think that's true. Your suggestion was probably just not seen. There is definitely the will to continue to support as many devices as we can, and to expand our device support to be as "universal" as we can be.

Nadahar:
@SubJunk I think I'm starting to get used to "Sami English" because I didn't even think about that 😆 I read "advertise" as "announce", and I understood the location to be on the UMS web site itself in conjuncion with the downloads. It might be that @Sami32 actually meant the forum, but as that didn't make sense to me I interpreted it as the web page.

I don't think @Sami32's point was where exactly those should be linked to, but the fact that we should inform our users that the bundled versions of this and that won't work under these circumstances/OS'es, but by replacing it with "link" it will work. Instead, users are given no such information, which means that they will probably download and try but never get it to work and just uninstall.

I also miss things @Sami32 says he has said previously "all the time", and I think it's caused by the fact that I don't really understand what he means most of the time. That said, there's also a *lot* of things that are ignored by you, or where you make some comment and then drop it. I think it's understandable if we've gotten quite "num" about things being ignored and end up just thinking: I tried at least, I've done "my part".

I know that I've been thinking that about at lof of subjects, to take a recent example it's the discussion about the problem with the double subtitles entries. I know very little about how the subtitle "logic" works, by choice (because I've seen enough of it to know I'd have to do a lof of refactoring to be able to get involved), which means I think it's natural that you and @valib are the main contributors to the discussion. I feel that I've tried to get a discussion going several times there already, but it simply doesn't happen. I've then concluded that there's nothing more I can do, you seem to accept that this bug and fundamental logical design problem is there.

Sami32:
@SubJunk Sorry for my bad english :flushed: 

As @Nadahar said, i meant that on the download and annoucement page for UMS, a note, annouce should be done about XP and old macOS specific limitations.

It's not only respect for users and program itself, as if it doesn't work for them they will say that this program doesn't work, but it seem just logic to me.
Eventually, a link to the freely available engine version, FFmpeg, working for their OS can be displayed; that doesn't cost anything and should be helpful for users and UMS program, IMO.
That was my two cents.

**Note**: I was not talking about "Universal" for devices, but codecs, and OS with a way to users to be able to fix it easily by themself if they wish to, or to easily dowload directly a working version for their OS.

SubJunk:
@Nadahar  I haven't been good over the last year at keeping involved in discussions I should have been in and I admit that, and I'm trying to get better at it. I have been overwhelmed by life and now I finally have more space to give to UMS. I hope I'm not too late to prevent you guys from branching off to start your own thing, because I think we can make more progress as a team and I don't think our goals are very different.

@Sami32 I will be happy to provide more information about those things in our announcements. Can you provide me a link to the version of FFmpeg you like for Windows XP?

Sami32:
@SubJunk 
For **Windows XP**, the lastest UMS version working without any issue is 5.5, AFAICR.
The latest official working FFmpeg version is `20160714-f41e37b`:
https://archive.is/o/hbWbs/https://ffmpeg.zeranoe.com/builds/win32/static/ffmpeg-20160714-f41e37b-win32-static.zip
The latest homemade build: (working also for SSE1 CPU)
http://abechin.sakura.ne.jp/sblo_files/k-tai-douga/ffmpeg/ffmpeg_bin_legacy-86841.zip
Other link:
https://sourceforge.net/projects/ffmpegwindowsbi/files/2016-08-12-v3.1.2/?SetFreedomCookie

**For MacOS:**


**Note**: I remember also having asked you to make the forum rules font a little bolder and blinking for 3 to 5 seconds, as it is proved that it will attract much more attention when dynamic.
Most of the users don't even see it, or are afraid to post their log with privacy injuries (though i'm near sure that most of them use Chrome or have Android smartphone LOL)

SubJunk:
Do you have a recommendation for which FFmpeg version to use for the Windows XP release?

Sami32:
@SubJunk Yes. That one:
http://abechin.sakura.ne.jp/sblo_files/k-tai-douga/ffmpeg/ffmpeg_bin_legacy-86841.zip

Nadahar:
@Sami32 Ehm - isn't that the very old one for those without SSE2? That doesn't really have to do with Windows XP 😉

Sami32:
@nadahar Check at the number, it is the most recent build working on all Windows version i guess, the only drawback is that it doesn't support HEVC, nor the latest supported official one,

If you've a better proposition, i listen you.

Nadahar:
@Sami32 I have no idea of the different FFmpeg versions and what is what, I'm just saying that there are *very few* people out there that needs one compiled without SSE2 support - and I assume that one built with SSE2 support will have some performance gain from using SSE2. If not, why would they use it in the first place?

Sami32:
@Nadahar You could just stick to UMS 5.5 or buy a new computer or find an other media server, build their own FFmpeg or cry.
That not the choice that miss.
They also can download an already made build that satisfy their need. That's not your choice, that only user choice.
They can have an old SSE5 version or an updated version that work also for old CPU.

Nadahar:
@Sami32 Sorry, I don't understand what you mean. Link to whatever version you want for me, I just tried to point out that SSE2 support and XP support are two different, unrelated things.

Sami32:
@Nadahar Nothing more than i've already posted.

Sami32:
@SubJunk OK, i understand only now what @Nadahar said.
So yes, i should have been more explicit in my links.
The first one is for Windows XP and the second one is for old CPUs.

So i prefered the second as you asked only for one link, as it will cover every users.
But the first one should cover at least 95% of the users with a speed enhancement, so i guess that it should be prefered.

Sami32:
@SubJunk You've mistaped the address. If you want only the general archive page, you need to type:
http://archive.is/hbWbs

SubJunk:
The download for that file doesn't work for me when I click it from your link

SubJunk:
I've added a new section of our announcements called "Compatibility Notes" that can be expanded http://www.universalmediaserver.com/forum/viewtopic.php?f=3&t=10450

Sami32:
@subjunk I don't if it your computer or the australian network restriction but it's working fine for me.
So, Google will be you best choice to find the FFmpeg version that i referenced to you:
ffmpeg-20160714-f41e37b-win32-static.zip

Nadahar:
@SubJunk I think you have to take a look at the spam on the forum, it's gone completely crazy now and it feels like there's a new post every 5 minutes. I've been banning spammers for a long time, but after the update of the forum the "one click ban" button is removed so banning is a lot of work now. That means that in addition to them being able to sign up like they want to, they keep spamming on the accounts they have already created.

I don't know if the reCAPTHCA implementation is flawed or if it's a problem with the CAPTCHA itself. I read something about a cookie exploit they could use to "always get the easy CAPTCHA", but when I read the developer help for reCAPTHCA v2 it seems like it shouldn't be that simple. There's a lot of communication back and forth with Google in verifying the validity etc, so I'm thinking that the implementation might not do everything it's supposed to. I looked at the phpBB extensions, and there seems like there are some good anti-spam extensions there as well that might help relieve this. What is certain is that the forum will quickly drown in spam as it is now.

SubJunk:
Yeah definitely, I've been doing mass bans several times a day lately. Will keep doing that until I have time to do a code change.

SubJunk:
@Sami32 I'm not in Australia and don't have a restricted network. This is the link I get the 404 on: https://ffmpeg.zeranoe.com/builds/win32/static/ffmpeg-20160714-f41e37b-win32-static.zip
Can you upload the file so I can download it?

Sami32:
@SubJunk The fact is that i can and you cannot...i leave you explain it.
Sorry, New Zeland, no offense.
Here an other download link: https://we.tl/3OUmLWS7vB

capture_08132017_140129

Nadahar:
@Sami32 It must be a worldwide conspiracy then, since Norway apparantly is a part as well. I can see the page you've taken a screenshot of, but when I click on the file you specify, I get a 404. If I pick one of the two on top though, it works fine.

Sami32:
@Nadahar Then for one time i'm lucky.
Yes, that work straight forward.

SubJunk:
@Sami32 thanks for uploading it, but that website starts downloading a file ending in `.exe` so I don't want to download that. I just confirmed that it will work if you attach the zip file in GitHub by attaching another one I had
[ffmpeg-20170605-4705edb-win64-static.zip](https://github.com/DigitalMediaServer/DigitalMediaServer/files/1221462/ffmpeg-20170605-4705edb-win64-static.zip)

Can you do that?

Sami32:
@SubJunk Sorry, i don't even understand what you asked me.

But as you already have the file, i think that you should be able to do what you asked me ?

Nadahar:
@Sami32 The file he has isn't the correct version, it was just an example of how you could upload it here. The file you've uploaded seems highly suspect as it's an ```.exe``` file, not an archive and there's no telling what running that can do.

Sami32:
@Nadahar Thanks. Btw, i only understood now what you was explaining before, i get thet website but the specific build link give me an error as well. I didn't even tried it, think that it should work. My bad.

@SubJunk Finally, here it is: (as a 7Z file with a fake .txt extension, to be able to post here directly)
[ffmpeg-20160714-f41e37b-win32-static.7z.txt](https://github.com/DigitalMediaServer/DigitalMediaServer/files/1222835/ffmpeg-20160714-f41e37b-win32-static.7z.txt)

Nadahar:
@Sami32 You can make ```.zip``` files with 7-zip as well, and they don't need the fake extension 😉

Sami32:
@Nadahar Using a ZIP extension will not make the size under the 10MB size :stuck_out_tongue:

Nadahar:
@Sami32 Look at the ```.zip``` @SubJunk posted, it's much more than 10 MB. Maybe they have changed their rule for ```.zip``` files or for all files..

Sami32:
@Nadahar I don't know, and don't really bother just make it work straight away, but as you already now it was 10 MB not long ago, for free Github account, and 25 Mb, whatever extensions, for some professional account (I don't knwo if they have different limit among the professional accounts).

Anyway, i'm not going to play with it anymore, so feel free to do it if you want.

SubJunk:
Thanks for the file @Sami32 
Can you please upload the win64 version too?

Sami32:
@SubJunk Sorry, i don't have it, don't forget that i use Windows XP 32-bit.
How many UMS users use Windows XP 64-bit ?

SubJunk:
@Sami32 when I used Windows XP it was the 64-bit version. I don't know how many we have on 64-bit. About 1% of the visitors to our website are on Windows XP though.

I have the `403a53c (20160617)` version of 64-bit FFmpeg since that's a version UMS had included.

Sami32:
@SubJunk Good, if you said that it work on WinXP 64-bit, then you have a working version for each one, as `403a53c (20160617)` will not work for WinXP 32-bit.

Nadahar:
Communication sure isn't easy 🤣 I must say that it give me some joy to see that you two has at least as much communication problems as me and @Sami32 - because then I know it's not just me. I can be quite frustrating some times, trying to find new ways to say the same thing in the hope that it will finally make it to the other end.

@Sami32: @SubJunk didn't say that he tested that version with 64-bit XP, he said that when he (once upon a time) used XP he had the 64-bit version.

When it comes to 64-bit Windows XP, this Windows update statistics from 2010 backs up my impression of the distribution: 
64bitvs32bitos_preview

At the time, very few computers had more than 4GB RAM, which was claimed to be the only reason to use a 64 bit OS. I've since discovered that it's not entirely accurate, the much larger *virtual* address space is a huge advantage in itself because every process can have its own address space instead of using a shared one. In any case, I believe the general consensus was that there were no point in using 64-bit Windows XP, and there's always the risk of added problem using new technology. 64-bit XP isn't "really XP" either, it turns out that 32-bit XP has Windows internal version number 501, while 64-bit XP has internal version number 502 - which is the same as Windows Server 2003. It therefore seems like 64-bit XP is really a "limited, rebranded" version of Windows Server 2003 and actually is a "generation newer" than 32-bit XP. That makes it hopeless to speculate what it supports and doesn't, the best source of information is probably that available for Windows Server 2003.

I've never had a copy of 64-bit XP, so I've never actually tried it myself and thus has zero experience with it.

SubJunk:
@Sami32 weird that Windows XP requires a newer version of FFmpeg than we already use.
I have created a branch for Windows XP, so now we use:

```
OS:              Build:
Windows Vista+   403a53c (20160617)
Windows XP       f41e37b (20160714)
```

Please test https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1370

Nadahar:
@SubJunk It doesn't require a newer version - that's all a misunderstanding. Windows XP requires a version that is built with something else than MSVC 2015 or later. In addition, parts that use [Direct2D](https://msdn.microsoft.com/en-us/library/windows/desktop/dd370990(v=vs.85).aspx) (if any) obviously won't work, so they need to be excluded from the build. As far as I can understand this isn't related to the version at all, but mainly which build tools are used.

This isn't a problem specific for FFmpeg though. Since Microsoft rolled out MSVC, although building for XP is still supposed to be supported, it's not by default and I'm not sure if enabling the "XP special settings" will produce something that works either. What I know is that any software built with MSVC 2015 or later tends not to work in Windows XP, so the "problem"/sabotage will apply to more and more software as developers upgrade.

Sami32:
Well, if i can add my 2 cents:
In fact it is possible to build for Windows XP under MSVC 2015, but as it's not configurate by default, most users forget about. It what have happened to MediaInfo between 7.81 and 7.91 i think.
For that, you have to use the `v140_xp` Platform Toolset rather than the default v140 Platform Toolset.

Nadahar:
@Sami32 I already linked to that in my previous post - if you'd checked the link. I have read stuff on forums that indicate that it's not always the case, although I can't be sure. That's why I'm reluctant to claim that it works with the compatibility target setting. Using them will make it *run* on XP, but I'm not sure that it will run without issues.

Sami32:
@Nadahar 
> if you'd checked the link

LOL :stuck_out_tongue: 
Well, obviously, for the FFmpeg case, some libraries need to be withdrawed as well, but is an other story.

SubJunk:
@Nadahar so I'm confused now, does the version of FFmpeg that @Sami32 uploaded work on Windows XP and our current one doesn't? Or do neither of them work? Or both?

Sami32:
Since you are working on it, i should win time asking where is the code to disable _tsMuxeR_ for anything other than transcoding to MPEGTS format ?

Nadahar:
@Sami32 I don't know, I bet it's spread around. What else is it used for?

Sami32:
`player.name()`
I don't really know if it is used for other thing, but i know that having the name of the video engine displayed in the #--TRANSCODE--# virtual folder when it doesn't support the format is just of no use, except confuse the users, IMO.
So, i'll check tomorrow how to do that.

Nadahar:
@Sami32 I haven't looked into the TRANSCODE folder logic much, but I thought it was supposed to only show those supported. But, wait, isn't that what @valib recently changed to fix a problem with his renderer?

Sami32:
OK, i think i've found the flaw in the code. The compatibility test is weirdly done.
I'll test it  tomorrow.

Sami32:
If you decide to finally upgrade the video engines :wink: you will see that the old _tsMuxeR_ used, i already deferred all subtitles to a recent FFmpeg version, you can download the latest _tsMuxeR_ here:
https://www.videohelp.com/software/tsMuxeR

It support the latest Mac OS version, the old one used stoppped at 10.4, and some subtitles and a little more formats.

Nadahar:
@Sami32 When I get around to looking at the engines themselves (this PR is just about "registering", testing and choosing them), I'm hoping to get rid of tsMuxeR alltogether. Since it doesn't seem like the author is going to release the source code and the project is dead, it's quite hopeless to use. It will probably be more and more problems as time goes by.

I was hoping that FFmpeg/Libav or some other updated software was able to do the muxing. If not, I'm wondering if muxing to TS really is necessary. I studied the TS format relatively closely some years ago, reading through the standard and I started to make a parser (it can parse which streams and types is in there) with the goal of extracting and converting the DVB subtitles. The TS format is quite complex, and isn't really meant for carrying a single media/set of streams. It is used in broadcasting, for example for sattelite television. The sattelites has something called transponders, which is basically separate radio senders on the sattelite (with their own frequencies etc.). Each transponder transmits only *one* TS stream, which carries multiple TV channels, radio channels, decoding information, EPG, decoder firmware upgrades etc. This is what the TS format is made for, as a "highway" for many different streams. 

The PS format is meant to carry just a single "program" or channel, and is much less complex. The two is still very similar though, PS is basically just a subset of TS. Still, it should be much less complex to mux to PS than to TS, and any decoder capable of decoding TS should also be capable of decoding PS. Because of this, I'm not sure there's a need to transcode to TS at all.

If FFmpeg/Libav or some other software is capable of transcoding to PS in a more "compatible" way, I can't quite see why that couldn't be used instead and TS simply be ditched alltogheter. As I've looked into the complexity of a TS stream, I fully understand why it is difficult to mux correctly. The equipment making the commercial TS streams are expensive and complex hardware devices made specifically for that.

Sami32:
Well, i'm still unsure about tsMuxer, its a good tiny tool that make TS remuxing very well.
AFAIK, it is maily used for PS3, but as you saw the japaneese user issue, some renderer only accept MPEGTS and are very strict on timestamp enforcement, as required in some DLNA profiles.

I've a FFmpeg patch for this, it has been done by a german some years ago but never got merged, but until i search and found a tool being able to do this well, i will suggest to leave it by now.
The main problem is with timestamps issues in TS files, AFAIK.
Other than TS are correctly handled by FFmpeg.

Nadahar:
@Sami32 Are you saying that there are renderers that accept TS but not PS? I find that hard to believe. I'd also think that the timestamps were the same for TS and PS, they are essentially the same but TS can do so much more that you don't need for a single media.

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1202#issuecomment-312481231

Nadahar:
@Sami32 I see that you're right. I don't quite understand why, but they must have dropped extending PS for some reason in later standards. As far as I can remember, it should be capable of doing "the same" from a technical point of view, but if the newer codecs aren't "defined" for PS it has become more or less useless. Too bad, I dislike tsMuxeR only because it's buggy and unmaintained.

Can't you get FFmpeg to fix it? 😇

Sami32:
Me, its the same for _MEncoder_, since i discovered all the issues and  unability to produce streamable MP4 :wink: Too old...

I didn't really searched how to do that yet, i'm at fine tuning my MP4 transcoding so that take me more time that i expected. Actually i make myself problem while touching at the DLNAResource file and searching later what part have done the break or issue :P

Sami32:
FFmpeg will certainly have fixed DVB Subtitles by the end of this years if not today, so that already good, isn't ? :wink: 
I planed to test it before Christmas.

Nadahar:
@Sami32 Who told you that MEncoder could produce MP4? I know it's broken for MP4, but that doesn't mean it's broken for everything else, although it isn't my favorite either as it's no longer developed.

I think your only choice for MP4 is FFmpeg. You should be very careful with changing things in DLNAResource, as that class is a nightmare of a mess that will break things *very* easily. So many other classes are subclasses of DLNAResource, which means that it can be very difficult to understand the consequences of changes.

While DVBSUB would be nice, it all depends on what it's transcoded to. I think the only format that can properly represent it is Blueray SUP files. Is that transcoding implemented? Any form of OCR "transcoding" will "fail" as I see it, the point with using those subtitles is their superiour layout/visual quality - which is lost if converted into text.

Sami32:
@Subjunk with his attempt to implement it, and you before with your attempt to use it for DVD streaming in Web browser. So when i understood that this engine was not able to make it MP4 streamable because it is too old and doesn't implemented fragmentation and moov atom code...

Yes, as i told you, some days ago they have fixed alpha subtitles handling and today the PGS encoder.
Enjoy :wink:

Nadahar:
@Sami32 It can't produce streamable MP4 - no doubt about that. But, UMS doesn't transcode *anything* to MP4 today, so the lack of MP4 ability doesn't mean it can't be used for anything.

Sami32:
In your fork or UMS fork,yes, but it does MP4 transcoding in mine :P
As i already said, yes, for now i just keep it for DVD, until Handbrake implementation.

Nadahar:
Review status: 0 of 121 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/main/java/net/pms/util/Version.java, line 744 at r9](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2#-KqEXRlAd2HOth7NL6ig-r9-744:-KuoxiQ-zaAdIipDqMyu:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/3cb40ea872aee1ef3ddc329a61b27db2317bb177/src/main/java/net/pms/util/Version.java#L744)):*
<details><summary><i>Previously, Sami32 (Samuel) wrote…</i></summary><blockquote>

Mistyping: `by by`
> by the LoadLibrary function.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2)*

Nadahar:
Review status: 0 of 121 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[.gitignore, line 8 at r15](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2#-Kvw_pOf7k5ussFF5H6t:-Kvw_pOf7k5ussFF5H6u:b-o0soab) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/84ca7c782784a81603bbcbd17ddbd0db7b01a079/.gitignore#L8)):*
> ```
> .idea/
> libmediainfo.dylib
> MediaInfo.dll
> ```

Note to self: Rebase this file out, as nothing is changed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2)*

Nadahar:
@sami32 I know you probably won't believe it, but I'm finished with this PR. Now we just need to find the bugs :stuck_out_tongue_winking_eye: 

---

Review status: 0 of 275 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2#-:-KxPb7JIHDvn2SlsnwHF:b-i7a7bg)*

Sami32:
I tested it this afternoon. That's was much better, no more epilepsy crisis :wink: 
But it should stay 1 to 2 seconds full red lightened before shade off, IMO.
If i remember correctly, it's about 1 or 2 value to change into a line of code ?

Nadahar:
@Sami32 It's easy to adjust the timing, but I'm not sure if it will be better. I can try it out.

It seems like there are bigger problems though. I just tested it on my XP and macOS VMs, and things didn't really work as intended. On the XP computer it reported some error for FFmpeg, but that might be "real". The fact that VLC wasn't installed wasn't ideal, as it reported it as "unknown error".

On macOS it's even worse, the only engine that was "green" was DCRaw. I'm not sure exactly what went wrong, but it seemed like both FFmpeg and MEncoder hung when the tests were run. It might be that I need to increase the time to wait for the tests to complete.

Have you seen any errors with the registration itself, or does it correctly find the engines you have installed/available?

Sami32:
Arrggg...correct. I installed a working _FFmpeg_ version and the result was the same: Error
I suspect that it could be related to Avisynth ? as it seem mix them and that i didn't installed Avisynth.

Nadahar:
@Sami32 I'm not sure what happens, I think the time might be too short. I've only given it one second to do the "test", it might not be enough on slow computers (like my VM's).

Sami32:
I also thought about a timing issue, but why only FFmpeg seem affected ?
```
TRACE 2017-10-27 03:23:01.718 [TPW "mplayer.exe" 1] Executing "C:\Program Files\Digital Media Server\win32\mplayer.exe dummy"
TRACE 2017-10-27 03:23:01.734 [Process Terminator] ProcessTerminator scheduled shutdown of process "mplayer.exe" in 300000 milliseconds
TRACE 2017-10-27 03:23:01.734 [Process Terminator] ProcessTerminator is waiting 299994 milliseconds
INFO  2017-10-27 03:23:01.781 [main] Checking transcoding engine AviSynth/FFmpeg
TRACE 2017-10-27 03:23:01.781 [main] Registering custom executable path for transcoding engine AviSynth/FFmpeg
TRACE 2017-10-27 03:23:01.796 [main] Executing "C:\Program Files\Digital Media Server\win32\ffmpeg.exe -version"
TRACE 2017-10-27 03:23:01.828 [Process Terminator] ProcessTerminator scheduled shutdown of process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" in 1000 milliseconds
TRACE 2017-10-27 03:23:01.828 [Process Terminator] ProcessTerminator is waiting 999 milliseconds
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1] Discarded 12 lines of process output:
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: MPlayer SB65-UMS (C) 2000-2016 MPlayer Team
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: Custom build by SubJunk, http://www.spirton.com
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: 
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: 	Scanning dir C:/WINDOWS/fonts
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: getch2: 6 can't get number of input events  [disabling console input]
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: 
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: Playing dummy.
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: File not found: 'dummy'
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: Failed to open dummy.
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: 
TRACE 2017-10-27 03:23:02.390 [TPW "mplayer.exe" consumer 1]   Process output: 
TRACE 2017-10-27 03:23:02.406 [TPW "mplayer.exe" consumer 1]   Process output: Exiting... (End of file)
TRACE 2017-10-27 03:23:02.406 [Process Terminator] ProcessTerminator unscheduled process "mplayer.exe" (2940)
TRACE 2017-10-27 03:23:02.406 [Process Terminator] ProcessTerminator is waiting 414 milliseconds
DEBUG 2017-10-27 03:23:02.828 [Process Terminator] Trying to terminate process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" (2832) since its allowed run time has expired
TRACE 2017-10-27 03:23:02.828 [Process Terminator] Attempting to stop timed out process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" (2832) with WM_CLOSE
TRACE 2017-10-27 03:23:03.031 [Process Terminator] Can't find any Windows belonging to process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" (2832), unable to send WM_CLOSE
TRACE 2017-10-27 03:23:03.031 [Process Terminator] Attempting to stop timed out process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" (2832) with TaskKill
TRACE 2017-10-27 03:23:03.187 [Process Terminator] Taskkill performed for process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" (2832)
TRACE 2017-10-27 03:23:03.187 [Process Terminator] ProcessTerminator is waiting 499 milliseconds
TRACE 2017-10-27 03:23:03.281 [Process Terminator] ProcessTerminator unscheduled process "C:\Program Files\Digital Media Server\win32\ffmpeg.exe" (2832)
TRACE 2017-10-27 03:23:03.281 [Process Terminator] ProcessTerminator is waiting for new tickets
DEBUG 2017-10-27 03:23:03.500 [main] Reading ffmpeg_executable_type: "Bundled" (default: null)
WARN  2017-10-27 03:23:03.515 [main] Transcoding engine "AviSynth/FFmpeg" is not available
INFO  2017-10-27 03:23:03.531 [main] Checking transcoding engine AviSynth/MEncoder
TRACE 2017-10-27 03:23:03.531 [main] Registering custom executable path for transcoding engine AviSynth/MEncoder
TRACE 2017-10-27 03:23:03.531 [main] Executing "C:\Program Files\Digital Media Server\win32\mencoder.exe -info:help"
TRACE 2017-10-27 03:23:03.531 [Process Terminator] ProcessTerminator scheduled shutdown of process "C:\Program Files\Digital Media Server\win32\mencoder.exe" in 1000 milliseconds
TRACE 2017-10-27 03:23:03.546 [Process Terminator] ProcessTerminator is waiting 998 milliseconds
```

Nadahar:
@Sami32 On my VM's not only FFmpeg is affected. It might vary how slow they are to start. One thing that makes the log a bit confusing is that at the same time as the tests are being run, the "fontconfig initialization" is running. Both FFmpeg and MPlayer is being run to generate fontconfig, and I've set the timeout for those to 300 seconds (as they might take some time). 

> It display an error message as a CTRL+C have been done, but that not the case, so something should have simulate it.

Where does it show that message? In the log or somewhere else? I've made it so that DMS tries to attach to the console of a running process and "press Ctrl + C". It's an attempt to shut it down "nicely" instead of killing the process. If it fails it will just move on and try other things, so that's no "problem" in itself.

Sami32:
The message appear into the GUI.

If its only a timing issue you will test it rapidely.

Nadahar:
@Sami32 Can you take a screenshot of the GUI message, because I'm not sure I understand what you mean?

Sami32:
Nothing to worry about, just a message that you have made to display, so you should have the same i guess.

Nadahar:
@Sami32 I haven't made any GUI message about Ctrl + C that I know of. It might be no problem at all, but at the moment I don't understand what it is.

Sami32:
capture_10272017_043446

Nadahar:
@Sami32 Thanks 👍 I hadn't seen that, but it the result of ProcessManager shutting the process down prematurely. I have to figure out what is going on, and my timeout is probably way too short.

Nadahar:
Increasing the timeout and fixing the file permissions for the executables on Linux made things work a lot better.

Nadahar:
It fixed the problems on macOS as well - and it reveals that the currently bundled version of tsMuxeR doesn't work - at least not in the 10.12 that I'm using: ```Segmentation fault: 11```.

Nadahar:
@Sami32 I've tweaked the light more like what I think you're suggesting. I think it gets quite annoying that way though, but I'd appreciate your feedback.

Nadahar:
@Sami32 Regarding AviSynth, I correctly get this message when AviSynth isn't installed:
```
Transcoding engine AviSynth/FFmpeg is not available since AviSynth could not be found.
```

Did you get some other error?

Nadahar:
@Sami32 Now I think this is finally ready. Any testing or review effort is very welcome. I've fixed all the issues I've found, and tested it on different platforms.

Sami32:
I checked at your 2 latest commits, so if you didn't have made more change into the others that i already reviewed that should be fine.
I will test it later.

Nadahar:
@Sami32 The last commits are mostly small stuff, but moving MediaInfo.dll might cause some issues in your "special configuration"?

Nadahar:
@Sami32 I might have rebased some very minor stuff into earlier commits, but nothing that should need "re-review".

Sami32:
Ok, but i hatdly can test it, as it look so nice that i just look at the icons :wink:
I just noticed one thing that i don't know if that should be addressed or not ?
When trying to run a scan and FFmpeg is defectuous, no advertising is done that the search engine is broken.

**Note**: Not like you, i never really noticed before now that shared folders can only be added one by one with the UMS code.

Nadahar:
@Sami32 Hehe, that's nice 🙄 

I'm trying to isolate things, I plan to refactor parsing and I will deal with that then. 

Do you think this is good to merge?

Sami32:
Hurry as you are i think that you will have already merged before my answer :tongue: 
But yes, that's great job :+1:

Nadahar:
Reviewed 28 of 106 files at r10, 5 of 94 files at r16, 43 of 82 files at r17, 11 of 26 files at r18, 2 of 4 files at r19, 6 of 9 files at r20, 2 of 9 files at r22, 12 of 16 files at r23, 12 of 14 files at r24, 142 of 144 files at r25, 10 of 11 files at r26, 4 of 4 files at r27, 5 of 5 files at r28.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/net/pms/util/Version.java, line 744 at r9](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2#-KqEXRlAd2HOth7NL6ig-r9-744:-KzC6lxF9bx2PMWTu_Cb:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/3cb40ea872aee1ef3ddc329a61b27db2317bb177/src/main/java/net/pms/util/Version.java#L744)):*

Done.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/2)*

PR #3, opened 2017-07-26
MIME types update

Sami32:
I'm the faulty one concerning WAV, as i searched to stick to the more "standard" one, but afterward the most widely supported should have been choosed.

About M4A, i suspect that 90% users need `audio/mp4` MIME type by default, because AAC is far more supported / used than ALAC and most renderers support it as well. :tongue: 
That was my subversive PR :wink:

That's said, that will permit to clean later, all the renderers from unecessary audio MIME types entries

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/3)

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -607,7 +607,7 @@ CharMap =
 #    mp2      MPEG Audio Layer II                           audio/mpeg
 #    mp3      MPEG-1 or MPEG-2 Audio Layer III              audio/mpeg
 #    mp4      ISOM/MPEG4 container, or MPEG-4 codec         video/mp4                          For Video files
-#    mp4      ISOM/MPEG4 container, AAC-LC/HE-AAC/ALAC      audio/mp4                          For Audio files
+#    m4a      ISOM/MPEG4 container, AAC-LC/HE-AAC/ALAC      audio/mp4                          For Audio files
 #    mpa      MPEG Audio                                    audio/mpeg
 #    mpc      MusePack                                      audio/x-musepack
 #    mpeg1    MPEG-1                                        video/mpeg                         Used in VCDs
@@ -640,7 +640,7 @@ CharMap =
 #    vp8      Google VP8
 #    vp9      Google VP9
 #    wavpack  WavPack                                       audio/wavpack
-#    wav      Waveform Audio File Format                    audio/x-wav
+#    wav      Waveform Audio File Format                    audio/wav
 #    weba     Webm Audio                                    audio/webm
 #    webm     WebM Video                                    video/webm
 #    wma      Windows Media Audio                           audio/x-ms-wma
--- a/src/main/java/net/pms/network/HTTPResource.java
+++ b/src/main/java/net/pms/network/HTTPResource.java
@@ -58,7 +58,7 @@ public class HTTPResource {
 	public static final String AUDIO_EAC3_TYPEMIME = "audio/eac3";
 	public static final String AUDIO_FLAC_TYPEMIME = "audio/x-flac";
 	public static final String AUDIO_LPCM_TYPEMIME = "audio/L16";
-	public static final String AUDIO_M4A_TYPEMIME = "audio/x-m4a";
+	public static final String AUDIO_M4A_TYPEMIME = "audio/mp4";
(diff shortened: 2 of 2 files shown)


Nadahar:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/3#-:-KpwXUPal7pz_fbeEKQr:bnfp4nl)*

Nadahar:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/3)*

Sami32:
I become suspicious ...we did argue about `audio/x-m4a` until i finally resigned without hope to make you understand, and now you approve it without any comment.

Hum hum hum... someone have usurped your identity or you'll discretly throw it in a hidden dustbin or refactor it :wink: 

Note: That could seem strange to you, but i finally get used to your comments, and also buyed a punching ball :rofl:

Nadahar:
@Sami32 So much of our discussions about mime-types has been because I've tried to say that codecs shouldn't have mime-types. I don't remember if I at one stage thought that M4A should have ```audio/x-m4a```, but now I know that DLNA use ```audio/mp4``` so I have no problem with that. That said, my long term plan isn't to use the current system anyway, I want to expand what's more or less implemented for images to audio and eventually video. That means using DLNA profiles, where the mime-type is dependent on the profile used and is clearly defined.

Sami32:
I share the same view, but i don't forget the SinkProtocolInfo limitations, by standard or weak implementation, that will always need a manual adjustement.
 And also that DLNA profiles have limitations that manual renderer's configuration can go beyond / fine tune, as i hope you saw in my framerate branch. Not talking about UPnP renderers.

Nadahar:
@Sami32 I'm not sure I share the worry about "weak" DLNA implementations. I realize that some implementations will be worse than others, but I don't think it's fair to assert that this is so bad before we've ever tried. UMS has never been anywhere near supporting DLNA, so the reason things work as well as they do is because the renderers aren't too picky. I believe that the combination of DLNA profiles and UPnP mime-types a renderer announces probably is close to what it can actually handle, and my goal is to not have to "define" anything manually for a renderer to use the supported profiles. Having the possibility to manually add or remove profiles to/from what the renderer announces should be possible, but I really doubt that's needed in most cases.

Sami32:
You seem misunsderstood me. I never asserted anything, and again less that it was bad.
Just, as you already now, it cannot be perfect, not your fault ;-) and will still need manual adjustmnents.

Nadahar:
@Sami32 The idea of "manual adjustments" undermines the whole point of DLNA. I'm not convinced that it's needed except in a few cases, but it's impossible to know before it's been tested.

Sami32:
For example, take the DMP, they will need manuel configuration right ?
I mean evy renderer not "printing" SinkProtocolInfo, for whatever reason, will depent on manual configuration.

Don't misunderstand me, using automatic configuration and profiles based on SinkProtocolInfo is the way to go. I'm just saying that a manual way, like it or not, should also be possible to circumvent such rare but possible cases.

Nadahar:
@Sami32 No, SinkProtocolInfo is just a part of the equation. If no protocolInfo information is available, a variety of "standard" profiles should be offered. Remember that it works both ways, the renderer is required to pick a profile it can play when it is offered several. DLNA also defines some mandatory profiles that should always be there regardless of SinkProtocolInfo.

But as I said, yes, the possibility to manually override should be there. I just want to move away from that being the standard, I believe that most renderers will work just fine without any manual configuration.

Sami32:
OK, sorry, i understand what you mean only now...
Seen as this, yes, it should be fine in more than at least 95% cases, IMO.

Yes, know, i look like a little paranoid but i take that from you LOL (i'm kidding of course).
Have that possibility, rassure me. After having tested it without any issues, or identified them, the manual part could eventually be withdrawed later.

Nadahar:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/3)*

Sami32:
OK, so let see if now it will remember my merging setting :wink:

PR #9, opened 2017-08-02
Add the minimum sampling value supported into the renderer configuration

Sami32:
Already tested by @ler0y on his Philips PFL renderer.

Some renderers have minimum supported values different from the DLNA standard.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/9)


Sami32:
I'm sure that you think that a correct DLNA implementation should taking care of it.
But as @ler0y tested for us, his renderer accepted AIFF as low as 6000 Hz, and that cannot be defined by any standard, AFAIK ?

Nadahar:
@Sami32 No, I was thinking about that I'm intending to implement a general ```Range``` class that can handle and parse ranges in a more versatile way, so that you could for example define: ```6000-48000``` or ```5000,7000-12000,18000,20000,24000-``` or something like that. I'm not done with the "research" part of it though, because I'm pretty sure I read a standardized range syntax somewhere a long time ago, but I can't find it now. There's a couple of things it need to handle that I don't know exactly how to do:
* Unlimited start (-8000) and unlimited end (8000-). Using the ```-``` is a problem because how do you tell a difference between a range with an unlimited start and a fixed negative value. The solution *might* be to use ```..``` or ```...``` instead, but I hate to "make up" standards if one already exist.
* Including or excluding. It should be possible to specify both including and excluding the boundries, so that you can either have a range where ```4000 <= x < 8000```, ```4000 < x <= 8000```, ```4000 < x < 8000``` or ```4000 <= x <= 8000```.

Again, the "challenge" is how to define the syntax so that most people understand it, and it's reasonably easy to parse. Decimal separators for numbers might also pose a challenge if ```,``` is used in that locale. ```;``` might be more suitable to separate the "sub ranges", but that makes it less readable.

In short, I haven't quite figured out how the formatting should be yet. If you can find any standards or standard practices for this, it would be helpful.

Sami32:
It's going farest that it ever intended. I don't think that more than 1% users will need that precision.

That's said, it'll be a perfect implementation :P though i've a tendancy to chosen what seem the more easy way fullfilling users requirements ;-)
I don't any problem to do that in the code, but knowing a little you, you saw that as a general part that could be used for any other part or program :wink: 
I'm not aware of such "standard", and i'll need to force myselft to search on that.

**Note**: It's funny that you talk about `,` parsing, because i think that doesn't work for french in the IP filter line.
But i guess that you was already aware of the same issue in the IP filter code.

Nadahar:
@Sami32 I've been thinking of that for a long time, and not about this parameter specifically. It would be great to have a general "range possibility" in many situations. I just know that I saw a standardized syntax a long time ago (and that I found it a bit over engineered/difficult to use and parse), but I can't find it now. I'm not saying that we *have* to use such a standard, but I'd like to know what's already standardized before making a decision.

I didn't know about any problem with ```,``` in the IP filter specifically, but I know that the code is full of that kind of bugs since such considerations hasn't been made in most of the code. The philosophy seems to be more that of "the shortest possible path to something that kind of works in the intended scenario while disregarding all unintended scenarios". That said, ```,``` isn't valid in IP addresses, they are defined to use the ```.```, unless you mean as a separator between multiple IP addresses.

There is a point with inclusion and exclusion, believe me I've learned that the hard way. In many instances it isn't, but some times you can't define the value "closest too". It applies to double/float values for example, where you'd have to define an indefinite number of decimals to catch anything below a given number: 9.99999999999999999 as an upper limit when you want to limit it to anything smaller than 10 won't, work, because 9.9999999999999999999999999999 will fail for example. It's the same problem with other types of comparisons using for example alphabetic sorting that is locale dependent. You can't know when you define it what letters come before a given letter. There are other more complex cases too, when custom ```Comparators``` are used where you have no idea of the defined ordering or where it might even change depending on other circumstances.

Sami32:
When you mind have the world in it, mine only have the sampling frequency parsing, so i'm too lazzy, not motivated at all, to search perfection.
Though i can understand, at least partially, your point of view, but no fun for me.
Remember, just a patchy guy ;-)

Nadahar:
@Sami32 I've had the need for it many times, so it's not like I'm just trying to make it as complicated as possible. As I said, I've been thinking of a general solution for this for a long time, at least a year I'd say.

Sami32:
How unlucky i'm, it's fallen on my PR :wink:

Sami32:
The conventional way, AFAIK, is `[-13,1-5,9,21]`

Nadahar:
Not that this is sufficiant, but as a reference this is what Maven use: http://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

Sami32:
To be honest i didn't even searched at it, i didn't saw any immediate or useful application.

But 2 days agao, without much will, i did looked at you naming convention in the purpose of range values parsing "super class".
1) `-13, 5, 12 - 42`. The point id to leave a space between the `-` and the [number](https://msdn.microsoft.com/en-us/library/ee414238.aspx) to differentiate it from a negative number.
2) `...` as in Git and littterature (meanig something like "and many others things")
3) `:` as in EXCEL. Ex. `A12-A27`

That the 3 ways, by preference order, that i founder to be able to parse also the negative values without issues.

Nadahar:
@Sami32 Yes, I think ```...``` probably is the best option. I'm more usure what to do with ```,```.

Sami32:
`- ` seem again more popular. Are you afraid that users forget to add space ?

The same principe could apply for the separator, if a space is before or after it should be considerer as a separator.
That's said i suspect that will be not very well followed.
So the Microsoft Email, CSV, Excel and consort choice can be followed as it seem by far been the most popular after the comma: `;` as a separator.
https://lockone.wordpress.com/2015/08/06/excel-now-using-semicolons-instead-of-commas-in-separating-formulas-fix-it-now/

Using an other separator could give again more confusion to users.

Sami32:
I just checked at your link. Woua, i never saw that before, that doesn't look natural to me.
Why not just use the natural writing ?
`1.0< x< 2.0` could be parsed, and more easy to write and read ?

Nadahar:
@Sami32 Space can't be used, it's used as a parameter separator many places, for example in the supported lines. That would mean that it wouldn't be read as part of the same parameter. I also think relying on space is "risky", is it wouldn't be obvious to anyone trying to modify existing parameters.

```;``` is a possibility, although slightly "ugly".

```1.0<x<2.0``` only makes sense to developers/programmers, and it would require us to define the name of the variable in some way. I'm not sure ```x``` would be a good fit universally.

Sami32:
Btw, i verified the IP filter, and i was wrong, it's working fine, so it was a bad remembering...

PR #10, opened 2017-08-06
FPS implementation in renderer's configuration

Sami32:
To be able to fine tune the renderer's configuration supported lines.

A DLNA profiles implementation will not be able to do that for DMP and UPnP renderers.

**Note**: Since i don't know yet how to clean the mess in my fork, so i pushed a clean version from here.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/10)

PR #12, opened 2017-08-07
Add 24-bit audio support

Sami32:
It seem that it affect "strangely" the branches on your fork, as i cannot even access to my dmr-ms" and ""24-bit" branchesfrom Git  !


Nadahar:
@Sami32 Sorry, I don't understand. It all depends on the remotes you define. Since it's a shared codebase, you can add remotes both for DMS and UMS and your forks of both. Then you can checkout branches from either of these places. If your brances are local on your computer, you can also add your local UMS repository as a "remote" so that you can have access to those branches.

Sami32:
As i told you, i use Github to see my code, i prefer it to Tortoise for having a global view.
Thank you very much for your cleaning help :smile: 

Sorry, seing all that mess in the history of my PRs make me upset and disgusted, that i preferered care for the garden and see a movie. More my fault than your, but it seemed to me like a Sysyph work to clean it everytime, or pushing each time a new clean version.
I guess that when i'll be more quiet i'll need to do an automatic script cleaning, and forget about the conflictuous ones. I cannot make code if i need to continuously battle with Git.

Nadahar:
@Sami32 You can't let the "GitHub history" get to you, it's very simple: It will show all commits that's different (in any way) from master as a part of "your PR". That means that all commits back to the oldest commit I changed will show up. However, I don't understand how it's a problem, when you look at the commits it's easy to see which commits that "actually belong to your PR". 

There is no more mess in the PR than it was, it's master that's changed, not the PR, and the only reason you see the long list of commits (and conflicts) is because of how GitHub has chosen to display it.

All I did with this PR was to cherry pick one commit onto the new master and push. It took me less than a minute, so all "the mess" GitHub showed was of no consequence.

The conflict that's here now is "real" though, this PR and #10 had changed the same code, so manual intervention is needed.

I really don't understand how you can get any "overview" by using GitHub. I use TortoiseGit exactly because it's so much easier to see the changes there. First of all, the font is way to big on GitHub, and the colors are hard on the eyes (too much contrast) so that if you "zoom out" to get more text on the screen it gets hard to read. In TortoiseGit that's not the case. Also, GitHub doesn't use the whole screen, which makes the problem even worse.

I think comparing code in unified diff is "too much work" - takes to much time, I much prefer side-by-side. But, I have to use unified on GitHub because side-by-side means that you have to scroll horizontally for every line you want to read. 

Finally, Tortoise lets you compare each file separately, and you can have multiple comparisons open at the same time, so that you can see what's changed in related files at the same time. In GitHub you have to keep scrolling up and down like a mad man, forgetting what was there in the meanwhile. Tortoise also shows all the text, not only what has changed, and lets you quickly jump between changes with Ctrl + arrow up or down.

If I had to use GitHub's diff for anything more than small changes I would cry, so if that's what you're doing I don't understand how you manage to get any overview at all 😉

Nadahar:
@sami32 I've rebased this on latest master and resolved issue. I also fixed a couple of bugs (you were setting bitdepth into mimetype), and I've made a "suggestion" commit for things that seems unfinished. It still looks that there's something missing before this will actually work..? :innocent: 

---

Reviewed 1 of 6 files at r1.
Review status: 1 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/12#-:-KvwlGtFTnVB2aqAVR9e:bc4zywl)*

Sami32:
I leaved it as unfiniched, because i didn't see it with a high priority.
For now, i'm turning in loop on DLNA, so in some days or weeks i will look at your work, never know, before that my mind get free.
I hope that you are not in a hurry, because i don't work under pressure...

Yes, some part has not being done yet, because i was not sure how to abord them.

src/main/java/net/pms/configuration/FormatConfiguration.java
```diff
@@ -91,6 +91,7 @@
 	public static final String JPEG = "jpeg";
 	public static final String JPEG2000 = "jpeg2000";
 	public static final String LPCM = "lpcm";
+	public static final String LPCM24 = "lpcm24";
```
Sami32: Thanks ;)
As i was thinking of adding LPCM8 as well, and it was still vague how i should handle all of them correctly in the DIDL and contentFeatures, when i get distracted by other stuff, as i use to be so easily...

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -672,6 +676,7 @@ CharMap =
 #               be assigned.
 #    n:         The maximum number of audio channels.
 #    s:         The maximum audio sample frequency (Hertz).
+#    abit:      The maximum audio bit depth supported.
```
Sami32: I realize that it's notthe best abbreviate name, as users could think of audio bitrate.

Maybe `abd` ? or `bd` or `d` ?

src/main/java/net/pms/encoders/FFmpegAudio.java
```diff
@@ -177,7 +177,7 @@ public synchronized ProcessWrapper launchTranscode(
 			cmdList.add("" + params.timeseek);
 		}
 
-		// Decoder threads
+		// Decoder threads - (Sami32) Why don't use "auto" instead and leave FFmpeg deal with it ?
```
Sami32: "auto" is set by default, AFAIK.
Only if the users choose to select the core number, this part of thread will be set.

**Note**: `auto` will not work with Zeranoe build, only it equivalent `0` will. The Linux build with `auto`, AFAIK, but since `0` work for all _FFmpeg_ versions...

src/main/external-resources/renderers/DefaultRenderer.conf
Nadahar: @Sami32 I was thinking the same thing, but couldn't come up with a good replacement. Maybe ```abd``` or ```ad``` are preferrable?

src/main/java/net/pms/encoders/FFmpegAudio.java
Nadahar: @Sami32 Probably, as long as the default is the optimal for most people it doesn't hurt to let them choose. I don't know enough about what is the best setting to use though, but it sounds like ```0``` would be the best default?

Nadahar:
@Sami32 I'm in no hurry at all, I just wanted to let you know that I had updated them.

src/main/java/net/pms/encoders/FFmpegAudio.java
Sami32: If set to zero, as by default, no such thread setting will be set into the FFmpeg command line, equal to set -thread 0, so...
Yes, FFmpeg automatically use it by default if no thread setting are set, for the input as for the outpout, AFAIK.

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: As i'm afraid that users think that anything beginning by "ab" could be related to audio bitrate, "ad" or "d" should be prefered, safer.

src/main/java/net/pms/encoders/FFmpegAudio.java
Nadahar: @Sami32 Isn't this how it's already working? I'm not sure I understand what you want to change..

src/main/external-resources/renderers/DefaultRenderer.conf
Nadahar: @Sami32 Yes, I think the ```a``` needs to be there to mark it as an audio only criteria, so ```ad``` might be the best then?

src/main/java/net/pms/encoders/FFmpegAudio.java
Sami32: Yes, so i've no plan to change anything ;)

Nadahar: Ok, so the comment can just be removed then..?

Sami32: Yes

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: It seem reasonnable to me, as any way far better than my first choice, that i finally find confusing.

PR #13, opened 2017-08-09
Chromecast 2 renderer

Sami32:
I add it here first, as i can have a cleaner file without unnecessary MIME type and add some parameter that are only in DMS.


Nadahar:
@Sami32 Before you put to much work into this, I'd just want to say that I wish to remove Chromecast support alltogheter. First of all, ChromeCast isn't UPnP AV or DLNA, so it actually isn't "related" to what we do. Second, I think all parts of the implementation is so bad that it basically needs to be completely reimplemented. Third, I don't really understand how you can use a ChromeCast with UMS/DMS anyway, since it doesn't support UPnP AV or DLNA. I've only seen the mention of using BubbleUPnP to push to it. I don't know if there's other ways as well, but I see it as a fundamental problem to get it right if that's the case because of the way UPnP remote control works. Unless there's something fundamental I'm missing here, there's no way to adapt what is offered to a third device (like a Chromecast), which means that it simply won't work if the DLNA implementation is improved.

PR #20, opened 2017-10-14
MediaInfo parse logger

Nadahar:
@Sami32 I separated this from ```RebaseableMaster``` so that we can work with this "in peace".

- [x] Use ```parseDuration()``` and ```Duration``` instead of ```getDuration()``` and ```Duration/String1``` - check for side effects.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/20)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -3540,9 +3540,9 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			configurationSpecificToRenderer.isThumbnailGenerationEnabled() &&
 			(renderer == null || renderer.isThumbnails())
 		) {
-			Double seekPosition = (double) configurationSpecificToRenderer.getThumbnailSeekPos();
+			double seekPosition = configurationSpecificToRenderer.getThumbnailSeekPos();
 			if (isResume()) {
-				Double resumePosition = (double) (resume.getTimeOffset() / 1000);
+				double resumePosition = ((double) resume.getTimeOffset() / 1000);
 
 				if (media.getDurationInSeconds() > 0 && resumePosition < media.getDurationInSeconds()) {
 					seekPosition = resumePosition;
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -75,7 +75,7 @@ public class LibMediaInfoParser {
 			// Set General
 			getFormat(StreamType.General, media, currentAudioTrack, MI.Get(StreamType.General, 0, "Format"), file);
 			getFormat(StreamType.General, media, currentAudioTrack, MI.Get(StreamType.General, 0, "CodecID").trim(), file);
-			media.setDuration(getDuration(MI.Get(StreamType.General, 0, "Duration/String1")));
+			media.setDuration(parseDuration(MI.Get(StreamType.General, 0, "Duration")));
 			media.setBitrate(getBitrate(MI.Get(StreamType.General, 0, "OverallBitRate")));
 			media.setStereoscopy(MI.Get(StreamType.General, 0, "StereoscopicLayout"));
 			value = MI.Get(StreamType.General, 0, "Cover_Data");
@@ -1005,7 +1005,10 @@ public class LibMediaInfoParser {
 
 	/**
 	 * Parses the "Duration/String1" format.
+	 *
+	 * @deprecated Parse "Duration" with {@link #parseDuration(String)} instead.
 	 */
+	@Deprecated
 	private static double getDuration(String value) {
 		int h = 0, m = 0, s = 0;
 		StringTokenizer st = new StringTokenizer(value, " ");
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -8,6 +8,7 @@ import java.util.Locale;
 import java.util.StringTokenizer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import javax.annotation.Nullable;
 import net.pms.configuration.FormatConfiguration;
 import net.pms.configuration.RendererConfiguration;
 import net.pms.dlna.MediaInfo.StreamType;
@@ -17,6 +18,7 @@ import net.pms.image.ImageFormat;
 import net.pms.image.ImagesUtil;
 import net.pms.image.ImagesUtil.ScaleType;
 import net.pms.util.FileUtil;
+import net.pms.util.StringUtil;
 import net.pms.util.UnknownFormatException;
 import org.apache.commons.codec.binary.Base64;
 import static org.apache.commons.lang3.StringUtils.*;
@@ -53,14 +55,6 @@ public class LibMediaInfoParser {
 		return MI.isValid();
 	}
 
-	public static void close() {
-		try {
-			MI.finalize();
-		} catch (Throwable e) {
-			LOGGER.debug("Caught exception", e);
-		}
-	}
-
 	@Deprecated
 	public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile, int type) {
 		parse(media, inputFile, type, null);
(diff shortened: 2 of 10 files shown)


Nadahar:
@Sami32 Ok, I've done what I was planning to do, so it's your to "fine tune" 😉 

If you think it's confusing why everything is twice, is that I've kept the methods for logging in "non-columns" as well in case that is useful in the future. There is two methods for each "part" of the logging, like ```logGeneral()``` and ```logGeneralColumns()```. Only the "Columns()``` versions are used now, but I'd like to maintain both versions just in case.

Sami32:
OK, i don't see the use by now but if you think that it could be used...
Yes, that redundancy don't make sens to me, no yet at least.

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -2601,7 +2601,17 @@ public final String getDidlString(RendererConfiguration mediaRenderer) {
 				if (player != null && media != null) {
 					// Note: Can't use instanceof below because the audio classes inherit the corresponding video class
 					if (media.isVideo()) {
-						if (mediaRenderer.isTranscodeToMPEGTS()) {
+						if (mediaRenderer.getCustomFFmpegOptions().contains("-f avi")) {
+							transcodedExtension = "_transcoded_to.avi";
```
Sami32: Also if i've added many enhanvcement since then, i think about make it modify the MIME type as well, but i didn't yet checked at it and see if it can be easily doable.

Nadahar:
@Sami32 The column based version is mostly useful to you with your narrow screen. If I for example want to parse a lot of data into a table, Excel sheet or just use it for regex searches, the other version will be better. It isn't much extra job to do it both places.

If you don't rebase but just adds the commits with the changes you do, I can do the same changed to the "non-column" versions. That way you only need to change it one place..?

Sami32:
I don't know, i'll do what i can, and you will take what you want/can from my commit.
When i'll do, i'll use my DMS fork. Ok, i'll need to remember to not rebase...

Nadahar:
@Sami32 You can just push to this branch/PR. This is only about this change. The other commits that's shown in this PR now is just because ```RebasableMaster``` haven't been pushed to master yet. Don't think about them, they're not "real". This PR is only the last commit.

Sami32:
I'll need to modify that one as well:
https://github.com/DigitalMediaServer/DigitalMediaServer/pull/20/commits/7cd7cae2073ba704ece50c4c98340a2732759003

Nadahar:
@Sami32 
> I'll need to modify that one as well:
7cd7cae

Not in this PR..? It's not related to the MediaInfo logging.

Sami32:
No. Anyway nothing hurry.

Nadahar:
@Sami32 If we keep this "clean" in that it only modifies the MediaInfo parsing, the resulting commit will be much easier to "move around" (cherrypick) to different branches and forks.

Sami32:
You can already cherry-pick 2 commits for the Rebaseable branch.
One is for an enhancement of the extensions commit and the other a verbose formatting for the ASF fix.

But as i said, few days ago i thought about making it modify the MIME type as well, so i'll not consider it as finish by now. Anyway, could any code be really considered as "finished" ? i don't think so.

Sami32:
https://github.com/DigitalMediaServer/DigitalMediaServer/compare/RebaseableMaster...Sami32:RebaseableMaster

Nadahar:
@Sami32 Hehe, "finished" is relative. I know about the commits to look at, I have them in the gitter "list". I'm trying to do the EngineRegistration GUI design part before looking at those and updating ```RebaseableMaster```.

Nadahar:
@Sami32 Now I'm really confused - don't they exist in ```PS4_SRT``` as well?

Sami32:
Ok, that perfect for me, take you time ;)
As i have redone it into DMS for in the Rebaseable branch, that should make things more easy for you.

Nadahar:
@Sami32 Ok. I'll see if I can figure it all out when I look at it 😉

Nadahar:
@Sami32 I rebased this PR on master to get rid of all the commits that didn't belong here. I also changed reference frames to ```int``` and fixed the tests.

Nadahar:
I also noticed that there is one unresolved "issue" with this PR that needs to be resolved before merge, and added it as a "task" to this PR. I don't know why I didn't complete it at the time.

Sami32:
:+1: 

Well, now that i just read about `byte` type, i realize that it was fine, as 16 should be the maximum value :flushed: 
So i hope that it will not use unecessary memory that much ?

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -183,28 +188,31 @@ public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile,
...
 							currentAudioTrack.setCodecA(FormatConfiguration.WMA10);
 						}
-						currentAudioTrack.setLang(getLang(MI.Get(audio, i, "Language/String")));
-						currentAudioTrack.setAudioTrackTitleFromMetadata((MI.Get(audio, i, "Title")).trim());
-						currentAudioTrack.getAudioProperties().setNumberOfChannels(MI.Get(audio, i, "Channel(s)"));
-						currentAudioTrack.setSampleFrequency(getSampleFrequency(MI.Get(audio, i, "SamplingRate")));
-						currentAudioTrack.setBitRate(getBitrate(MI.Get(audio, i, "BitRate")));
-						currentAudioTrack.setSongname(MI.Get(general, 0, "Track"));
+						currentAudioTrack.setLang(getLang(MI.Get(StreamType.Audio, i, "Language/String")));
+						currentAudioTrack.setAudioTrackTitleFromMetadata((MI.Get(StreamType.Audio, i, "Title")).trim());
```
Sami32: https://github.com/MediaArea/MediaInfoLib/issues/680

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -183,28 +188,31 @@ public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile,
...
+						currentAudioTrack.getAudioProperties().setNumberOfChannels(MI.Get(StreamType.Audio, i, "Channel(s)_Original"));
```
Sami32: https://github.com/MediaArea/MediaInfoLib/issues/680

Nadahar:
@Sami32 Byte itself is fine. But, because modern CPU's are optimized for integers (long actually for 64 bit CPUs) they will be used as that anyway by the CPU, so there's nothing to save while processing it. In memory it does use 3 bytes less than an integer, but 3 bytes isn't really something to worry too much about. When we're talking big arrays it obviously matters and byte should be used when appropriate, but for a single variable it really doesn't matter. 

The main reason I like to stick to integers most of the time in Java though is Java itself. Whenever you write a number in the code (for example ```if (byte < 13)```), the number itself (13) is an ```int```. That means that to do that comparison, Java will first convert the ```byte``` to an ```int``` and then compare. It's better to be able to compare directly instead of keep converting them for ever test. To avoid the conversion you actually have to write ```if (byte < (byte) 13)``` or ```if (byte < 13b)```. That's just to much clutter and things to remember for me. Thus it's easier to just stick to ```int``` when doing stuff where it will do the job.

The last point is Java's lack of unsigned types. That means that a ```byte``` can never be more than 127, and if you write ```if (byte > 127)``` that will never be true even if you set ```byte = 150```. Java won't complain if you do that, it will instead assign the value -106 to "byte", and thus it won't be true. That's the really scary stuff using ```byte```, since in the rest of the world (outside Java) a byte is a number between 0 and 255 (unsigned) so it's common that a byte will have a value above 127.

The reason is that 150 in binary is ```00000000000000000000000010010110```. When converting this to a byte, all but the last 8 bits will be discarded and the byte value will be ```10010110```. That in itself is correct, but since Java only knows of "signed bytes" it won't read the first bit as having value 128, it will read it as a minus sign.... 😢

Sami32:
> The main reason I like to stick to integers most of the time in Java though is Java itself. Whenever you write a number in the code (for example if (byte < 13)), the number itself (13) is an int. That means that to do that comparison, Java will first convert the byte to an int and then compare. It's better to be able to compare directly instead of keep converting them for ever test. To avoid the conversion you actually have to write if (byte < (byte) 13) or if (byte < 13b). That's just to much clutter and things to remember for me. Thus it's easier to just stick to int when doing stuff where it will do the job.

Good, now i feel better :wink:

Nadahar:
@Sami32 Yes, I didn't *just* do it because of you comment, but because I agreed as well 😉 

One thing I forgot to mention above is that once you have converted 150 into a byte (and gotten -106), it won't help to convert it back to ```int```. Because of the rules of widening negative integers, it will be converted back to ```11111111111111111111111110010110``` (which is -106 as ```int```), so the "damage" is permanent unless you do ```byte & 0xff```. That will correctly turn it back to ```00000000000000000000000010010110``` because you do a binary AND between ``````11111111111111111111111110010110``` and ```00000000000000000000000011111111```.

This problem is true for all primitive numbers in Java (byte, short, int, long), but the problem is biggest with byte because that is usually considered unsigned while the others are usually considered signed. You usually have types like ```uint``` to represent unsigned integers, so a normal "int" is signed "by convention".

When you add all this confusion together with Java's auto type casting, it's easy to see a lot of stuff that you don't want can happen - which is why I prefer to be as explicit as possible and avoid auto casting and boxing when I can.

Nadahar:
@Sami32 I think this is ready now, do you have any objections?

Nadahar:
Reviewed 1 of 4 files at r2, 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/20)*

Sami32:
I'll have to take the time to look at the code as i forgot everything since then.
So if you are not in hurry...

PR #26, opened 2017-11-12
Default shared folders

Nadahar:
This PR creates implements the concept "default shared folders" as opposed to the previous "share all drives" practice. The default shared folders are platform dependent, and is for example ```Music```, ```Pictures```, ```Videos``` and more on Windows.

This also removes the code from https://github.com/UniversalMediaServer/UniversalMediaServer/pull/140 since this caused logical flaws and bugs in the shared folders handling.

Remaining tasks:
- [x] Implement enumeration of default shared folders on macOS
- [x] Implement enumeration of default shared folders on Linux
- [x] Redesign GUI solution

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/26)


Nadahar:
@Sami32 This isn't yet implemented for all platforms, but it is finished for Windows. It also has some code that only runs on Windows XP, so it would be nice if you could test it.

src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -2650,28 +2657,282 @@ private static String listToString(List<String> enginesAsList) {
...
+		if (foldersString == null || foldersString.length() == 0) {
+			return folders;
+		}
+		String[] foldersArray = foldersString.trim().split("\\s*,\\s*");
+
+		for (String folder : foldersArray) {
+			/*
+			 * Unescape embedded commas. Note: Backslashing isn't safe as it
+			 * conflicts with the Windows path separator:
+			 * http://ps3mediaserver.org/forum/viewtopic.php?f=14&t=8883&start=250#p43520
```
Sami32: This link doesn't exist anymore

src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -2650,28 +2657,282 @@ private static String listToString(List<String> enginesAsList) {
...
+		String listSeparator = String.valueOf(LIST_SEPARATOR);
+		ArrayList<File> tmpSharedfolders = new ArrayList<>();
+		ArrayList<File> tmpMonitoredFolders = new ArrayList<>();
+		for (Vector rowVector : tableVector) {
+			if (rowVector != null && rowVector.size() == 2 && rowVector.get(0) instanceof String) {
+				String folderPath = (String) rowVector.get(0);
+				/*
+				 * Escape embedded commas. Note: Backslashing isn't safe as it
+				 * conflicts with the Windows path separator:
+				 * http://ps3mediaserver.org/forum/viewtopic.php?f=14&t=8883&start=250#p43520
```
Sami32: same same

src/main/resources/i18n/messages.properties
```diff
@@ -434,6 +433,10 @@ RendererPanel.8=Specify a file name
 RendererPanel.9=Select a reference file
 RendererPanel.10=Name
 RendererPanel.11=Address
+SharedFolders.ConfirmRemove=This will remove %d folders. Are you sure?
```
Sami32: A whitespace before all the asking mark "?" should be more readable, IMO:
`SharedFolders.ConfirmRemove=This will remove %d folders. Are you sure ?`

src/main/java/net/pms/newgui/NavigationShareTab.java
```diff
@@ -274,16 +244,16 @@ public JComponent build() {
...
-			builder.add(fullyPlayedAction,                                         FormLayoutUtil.flip(cc.xyw(4,  29, 3), colSpec, orientation));
-			builder.add(fullyPlayedOutputDirectory,                                FormLayoutUtil.flip(cc.xyw(9,  29, 2), colSpec, orientation));
-			builder.add(selectFullyPlayedOutputDirectory,                          FormLayoutUtil.flip(cc.xyw(11, 29, 2), colSpec, orientation));
+			builder.add(fullyPlayedAction,                                          FormLayoutUtil.flip(cc.xyw(4,  29, 3), colSpec, orientation));
+			builder.add(fullyPlayedOutputDirectory,                                 FormLayoutUtil.flip(cc.xyw(9,  29, 2), colSpec, orientation));
+			builder.add(selectFullyPlayedOutputDirectory,                           FormLayoutUtil.flip(cc.xyw(11, 29, 2), colSpec, orientation));
 
-			builder.add(builderSharedFolder.getPanel(),                             FormLayoutUtil.flip(cc.xyw(1, 31, 12), colSpec, orientation));
+			builder.add(sharedPanel,                       						    FormLayoutUtil.flip(cc.xyw(1, 31, 12), colSpec, orientation));
```
Sami32: What are the formatting rules here ?

src/main/resources/i18n/messages.properties
Nadahar: @Sami32 That's syntactically wrong, it should be in the French translation only: https://www.quora.com/Why-do-French-speakers-place-a-space-before-exclamation-points

src/main/java/net/pms/newgui/NavigationShareTab.java
Nadahar: I have no idea :P I just aligned it to the others, but I should use spaces instead of tabs. Will fix.

src/main/resources/i18n/messages.properties
Sami32: Wow, i didn't knew that the french was the only one doing that !
Though your link is not right, we always add a white space after any ponctuation mark, but not after all of them, independently of even or odd strange story write there.
Never before `;` and i saw the 2 for `:`

Sami32:
Where is the code for XP ? and what was the need, the GUID ?

src/main/resources/i18n/messages.properties
Nadahar: @Sami32 I don't really want to know the French syntax - it's enough for me to know that, as usual, the French has to make their own rules that's different from those of everybody else. Just look at Citroen, it took many years before they even managed to get the indicator signal switch in the right place, I can't say I'm surprised 😛

Nadahar:
The code for XP is the CSIDLs, as XP doesn't support the GUID system.

Sami32:
OK. Thanks

src/main/resources/i18n/messages.properties
Sami32: :stuck_out_tongue:

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -263,53 +276,174 @@ private void scan(DLNAResource resource) {
...
+					if (permissions.isBrowsable()) {
+						return folder;
+					}
+					LOGGER.warn("Insufficient permissions to read default folder \"{}\"", csidl);
+				} catch (FileNotFoundException e) {
+					LOGGER.debug("Default folder \"{}\" not found", folder);
+				}
+			}
+		} catch (Win32Exception e) {
+			LOGGER.debug("Default folder \"{}\" not found: {}", csidl, e.getMessage()); //TODO: (Nad) Enum
```
Sami32: I think that you already have done CSIDL Enum ?

src/main/java/net/pms/newgui/NavigationShareTab.java
```diff
@@ -915,12 +916,12 @@ public void actionPerformed(ActionEvent e) {
 		builderFolder.add(scanButton, FormLayoutUtil.flip(cc.xy(5, 3), colSpec, orientation));
 		scanButton.setEnabled(configuration.getUseCache());
 
-		File[] folders = PMS.get().getSharedFoldersArray(false);
-		if (folders != null && folders.length > 0) {
-			File[] foldersMonitored = PMS.get().getSharedFoldersArray(true);
+		List<File> folders = PMS.get().getSharedFolders(false); //TODO: (Nad) Here
```
Sami32: I don't understand what had to be done here ?

src/main/java/net/pms/dlna/RootFolder.java
Nadahar: @Sami32 You're not reviewing the up-to date code, that message in no longer there.

src/main/java/net/pms/newgui/NavigationShareTab.java
Nadahar: @Sami32 Nothing, it was a "bookmark" for me to easily jump to that location. It's not there in the latest code.

Sami32:
Doesn't work on XP, it hang on the splash screen.

Nadahar:
@Sami32 Ok, thanks. It could be a JNA issue. I'll have to test on my XP VM.

Sami32:
That happen in the beginning of the DMS init, as i cannot even get a trace from _JavaVM_.

Nadahar:
@Sami32 It works fine on my XP VM. Have you done any tricks with JNA? You shouldn't have to replace anything, the correct JNA version should be used.

I built it with Maven and ran the jar manually with no problem. I also built and ran it with Eclipse with no problem. I can't quite understand why it doesn't work for you. Did you fetch first?

Sami32:
No tricks, only clean install.
Ok. So that's strange.

Nadahar:
@Sami32 What exactly did you do? Did you build with Maven and then run the installer and the run the installed version?

Sami32:
Yes

Nadahar:
@Sami32 That's the one thing I didn't do. Do you get the same problem if you run the JAR directly? If you use "java.exe" this also has the benefit of showing you the log from the very start:

```java -jar dms.jar```

**Comment by Sami32** [comment]
As you know i always use that way. I should have remembered "your" way though.
Error come without any surptise from initialization:
> `use_default_folders`  doesn't map to a Boolean object

**Comment by Nadahar** [comment]
Review status: 0 of 21 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/java/net/pms/configuration/PmsConfiguration.java, line 2818 at r1](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/26#-KykNbrV02bPuRsqUCzo-r1-2818:-KypZIwHo16q3-W4b1l_:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/54cc941c0eadeaf20e239e5849792691ccb12d37/src/main/java/net/pms/configuration/PmsConfiguration.java#L2818)):*
<details><summary><i>Previously, Sami32 (Samuel) wrote…</i></summary><blockquote>

This link doesn't exist anymore
</blockquote></details>

Done.

---

*[src/main/java/net/pms/configuration/PmsConfiguration.java, line 2894 at r1](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/26#-KykNbrV02bPuRsqUCzo-r1-2894:-KypZHVAja6R7GDyPqM1:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/54cc941c0eadeaf20e239e5849792691ccb12d37/src/main/java/net/pms/configuration/PmsConfiguration.java#L2894)):*
<details><summary><i>Previously, Sami32 (Samuel) wrote…</i></summary><blockquote>

same same
</blockquote></details>

Done.

---

*[src/main/java/net/pms/newgui/NavigationShareTab.java, line 251 at r1](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/26#-KykNbrV02bPuRsqUCzx-r1-251:-KypZgZmJY_kTHGh-dex:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/54cc941c0eadeaf20e239e5849792691ccb12d37/src/main/java/net/pms/newgui/NavigationShareTab.java#L251)):*

I have no idea :P I just aligned it to the others, but I should use spaces instead of tabs. Will fix.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/26)*

**Comment by Nadahar** [comment]
@Sami32 Do you have a bit more of the log, a stack trace for example? I don't quite understand why this happens.

**Comment by Sami32** [comment]
If you like, i didn't saw the need to but if it can help you:
capture_11132017_164434

**Comment by Nadahar** [comment]
@Sami32 Ahh, you need to run:
```java -jar target\dms.jar trace```

You don't get the trace log, so no strack trace.

I've made the parsing a bit more robust though, so try the latest version.

**Comment by Sami32** [comment]
Ok, but at init time you couldn't expect much trace ?
You have fixed it :+1: 
capture_11132017_162713

The one you saw is only because my firewall block any new instance connection not yet configured.

**Comment by Nadahar** [comment]
@Sami32 Trace is as useful during initialization as everywhere else. Does it work for you with this fix?

**Comment by Sami32** [comment]
>You have fixed it :+1:

Yes.

Why do you have set this new option "Customize folders" ?
I find the visual effect not nice, and "eating" place for users having a little screen ;)
Using it the way as before should be better, IMO.

**Comment by Nadahar** [comment]
@Sami32 I agree that the layout isn't great, but I decided to not put a lot of time into finding a good place to put it. Suggestions are welcome.

The reason for the option is the be able to choose between using the default shared folders or not. How is it better the way it was before, when the default was just to share everything? I really consider the previous behaviour as a huge bug, it's not what people expect DMS to do and it will lead to some really crazy scan times, not to mention potential errors for network attached or removable resources.

The reason I made it a choice is that I dislike when application does things "as it think is appropriate" without informing the user. That's "the Apple way", and I despise it. This way users can actually choose to share nothing without DMS overriding their decisions because it think it "knows better". It's also handy that you can copy the default shared folders into your custom shared folders as a starting point, where you can then add or delete folders as you wish.

The question is if enumerates your folders correctly if you enable "use default"?

**Comment by Sami32** [comment]
No error, scan on its way.

You misunderstood me. I mean that it will be better, IMO, if this option was removed from the GUI.
And users could, like before when that had already configured their folders, modify them the "normal" way. Do i'm more clear ?

**Comment by Nadahar** [comment]
@Sami32 No, I don't understand you. Users can still modify the folders the "normal" way. "Customize shared folders" only default to false if no folders are configured. That means that for configuration where folders are already defined, it should default to true and everything work as normal. Isn't that what happens?

There's no reason to do actually do a scan, I haven't fixed the restart bug so it will still scan the wrong folders if you changed this without restarting (I think). Looking on the paths of the default folders should be sufficient, you should see if they are properly resolved to the correct locations.

**Comment by Sami32** [comment]
Well, some issue seem appeared with folders having "special" character()s.
They are added many times:
```
INFO  2017-11-13 16:50:28.562 [main] A tiny cache admin interface is available at: http://192.168.1.29:5001/console/home
TRACE 2017-11-13 16:50:28.578 [main] Adding new child "Recently Played" with class "Playlist"
TRACE 2017-11-13 16:50:28.578 [main] Adding new child "New Media" with class "MediaMonitor"
TRACE 2017-11-13 16:50:28.593 [main] Adding new child "Ma musique" with class "RealFile"
TRACE 2017-11-13 16:50:28.593 [main] Adding new child "Mes images" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Forcing short path name on C:\Documents and Settings\All Users\Documents\Mes vidéos
TRACE 2017-11-13 16:50:28.609 [main] Adding new child "Mes vidéos" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Forcing short path name on C:\Documents and Settings\All Users\Documents\Mes vidéos
TRACE 2017-11-13 16:50:28.609 [main] Adding new child "Bureau" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Adding new child "Ma musique" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Adding new child "Mes images" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Forcing short path name on C:\Documents and Settings\Philippe\Mes documents\Mes vidéos
TRACE 2017-11-13 16:50:28.609 [main] Adding new child "Mes vidéos" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Forcing short path name on C:\Documents and Settings\Philippe\Mes documents\Mes vidéos
DEBUG 2017-11-13 16:50:28.640 [main] Added file watch at C:\Documents and Settings\All Users\DigitalMediaServer: C:\\Documents and Settings\\All Users\\DigitalMediaServer\\WEB.conf
DEBUG 2017-11-13 16:50:28.812 [main] Sending BYEBYE...
```

**Comment by Nadahar** [comment]
@Sami32 I'm not sure I understand the log above, what folders are listed in your "shared folders" table when "customize" if off? You can select the whole table with Ctrl + A, and then Ctrl + C it and Ctrl + V it here like this:

C:\Users\user\Desktop\testObjects | true
-- | --
D:\Movies | true
E:\data\Sourcecode\Java\Media Sample Video | true
E:\data\Sourcecode\Java\Media Sample Audio | true
E:\data\Music\MP3 | true

**Comment by Sami32** [comment]
```

C:\Documents and Settings\All Users\Documents\Ma musique | true
-- | --
C:\Documents and Settings\All Users\Documents\Mes images | true
C:\Documents and Settings\All Users\Documents\Mes vidéos | true
C:\Documents and Settings\Philippe\Bureau | true
C:\Documents and Settings\Philippe\Mes documents\Ma musique | true
C:\Documents and Settings\Philippe\Mes documents\Mes images | true
C:\Documents and Settings\Philippe\Mes documents\Mes vidéos | true
```

**Comment by Nadahar** [comment]
@Sami32 It seems like they are only listed once. What is confusing is probably that shared folders are only named by their folder name, which is the same for the "All users" and the "Personal" versions.

```C:\Documents and Settings\All Users\Documents\Ma musique``` and ```C:\Documents and Settings\Philippe\Mes documents\Ma musique``` will both be shown as ```Ma musique``` when browsing. I considered if I should make some solution to this (adding more of the path if some had the same name), but I decided to try to restrict myself to the task at hand. That's not really unique to the default folders, if you have multiple shared folders with the same folder name, there will be no way to tell them apart when browsing.

**Comment by Nadahar** [comment]
@Sami32 It seems to me like it's working for you, but you just don't like the way it's working 😉 

I assume that the layout is your biggest issue, and maybe that it's slightly confusing? Do you have any suggestions for how it can be improved?

**Comment by Sami32** [comment]
Ok, i don't understand.
```
TRACE 2017-11-13 16:50:28.609 [main] Forcing short path name on C:\Documents and Settings\All Users\Documents\Mes vidéos
TRACE 2017-11-13 16:50:28.609 [main] Adding new child "Mes vidéos" with class "RealFile"
TRACE 2017-11-13 16:50:28.609 [main] Forcing short path name on C:\Documents and Settings\All Users\Documents\Mes vidéos
```
If it is fine, you can merge it then.

I don't really mind for what i tried to explain you without success, i will try to patch it later after your merge to fit my taste.
The layout is fine for me, although i did planed with no ETA of doing some folder deeper level  information folder while scanning, but it is not related to this PR.

Nadahar:
@Sami32 I don't understand what you don't understand. The "forcing short pathname" system is something I've never looked into and I don't know why it's done or why it's logged. I have no idea why that's done twice, but it looks like the folders are actually just added once..?

The "short pathname" function converts "long" Windows filenames to their DOS era 8.3 counterparts (where a filename could only be 8 characters + a 3 character extension with no spaces and no unicode support). I'm guessing that's a hack done to circumvent a problem instead of solving it, either related to unicode or spaces. This is something that is done inside ```DLNAResource``` AFAICR and isn't related to this PR.

As to your above explanation I didn't understand it, so please try to explain a bit more. Did you simply want to remove the "customize folders" option? How should the user be able to specify custom folders then?

Sami32:
Aahh, sorry i didn't realized that the fact to add it 2 times was not coming from your PR :flushed:

Sami32:
The fact that nothing anymore is displayed when you have delete all the shared folders could be an issue to me, IMO. As users could not understand that it imply the full scan of his computer

Nadahar:
Hmm, I found something bad myself. The "customize" checkbox seems to register clicks wherever you click on that line. That's very confusing, and should be fixed.

Sami32:
Sorry, it was only a wrong guess, as i didn't tested it. You are right, it is really empty ;)
Never mind, just remove that line ;P

Nadahar:
@Sami32 I don't understand what you mean. If I remove the checkbox, this won't work. How are users supposed to switch between the default folders and the customized ones without it?

Sami32:
The default folders are automatically set by default.
If users want modify it they just have to use the ADD or DELETE buttons that should automatically trigger the custom mode.

Of course, it only my point of view.

Nadahar:
@Sami32 Ok now I think I understand what you mean. It's closer to how it works without this PR, but I think there are some problems (which is why I chose another logic):

* When in "default mode", should the actual folders be shown so that users can see what is shared, or should if only display some nonsensical text like "default", "all media", "all drives" or similar. I found it much more clear to display the actual folders.
* Assuming that displaying the folders is the correct choice, it's very confusing to users that if you add one folder, all the others will disappear and only that one folder remain. Likewise, it's very confusing that it doesn't work to remove one of the listed folders.
* If it was made so that adding or removing a folder from the default folders made a "copy" of the default folders and then added or removed folders to/from that, users might find it annoying to have to delete a lot of folders (on newer Windows versions for example there are more standard folders). On the other hand it might be that the user want to share most of these, but make some changes without having to browse to the location of all those folders. As a result I made it so that you can choose if you want to copy the default folders into your "custom" folders.

To make it easier to work with the shared folders table, I also made it so that it's possible to remove multiple folders at the time and so that it will insert new folders at the selected line if one is selected.

What's your thoughts about the considerations above? Do you think it would be logical if "add" removes all entries and remove isn't possible?

Sami32:
- As you did, display they name and location, is perfect.
>- Assuming that displaying the folders is the correct choice, it's very confusing to users that if you add one folder, all the others will disappear and only that one folder remain. Likewise, it's very confusing that it doesn't work to remove one of the listed folders.

I don't understand this.
When you decide to add one or more folders, or to remove some, including the default ones, the untouched will stay.
- I don't know about what number of folders you are talking about, but as it avoid the issue of loosing time and resource to sacan all the computer, i don't see the annoyance of delete 6 folders, i suspect that noobs will not even remove them, as i said that not the kind of things that you do often, IMO. 

>To make it easier to work with the shared folders table, I also made it so that it's possible to remove multiple folders at the time and so that it will insert new folders at the selected line if one is selected.

I see. But i don't really see any use of it, as the user know where are the folders that he need, and also considering that a total noobs did remove it, a tooltip advertising that an empty shared folder will set the default ons should be suffisant, IMO. That should not the kind of operation that users will do often ? I could be wrong though...

Nadahar:
@Sami32 I was simply thinking of the case where you want to remove the last shared folder and add another one - but as soon as you remove the last folder (before you have added the new) suddenly 15 "default" folders are added. I think I would find this pretty annoying, which is why I decided to allow to have no shared folders. It doesn't otherwise make any sense in having no shared folders, but for me at least it's natural to remove the old one first and then add the new one - without a lof of folders being added "by magic" in the meanwhile.

Nadahar:
@Sami32 Another thing to note is the idea that users "know where the folders are". Many don't. There are a lot of people using computers that don't even understand what files and folders are. They always use the default locations and names, and think those open or save dialogs are pure voodoo that should be closed as soon as possible.

This way, they can always get back the default folders simply by clicking the checkbox, and at the same times enabling the default folders won't overwrite their configured folders.

Sami32:
Ok, i see your point.
The default shared folder could be added after some time being empty but that could give weird result depending of the situations.
So using REMOVE button on an empty shared folder could be used to set the default ones ? or a double-click on the epmty shared folder ?

Nadahar:
@Sami32 That would be possible, but I'm not sure it would be very intuitive. I think you'll thing anything is a good idea as long as you get rid of the checkbox? 😝 

You don't have a better idea for where to place it or anther way to switch beween "default" and "manual" shared folders instead? I agree that it's ugly and ruining the layout, I just haven't found a way that's good.

Sami32:
Yes ;P
Well, i don't think it should be showed at all.
I will search to see if i can find a better switch for it.

Sami32:
Well, i've one, but i don't know if it is possible.
Can the add button have 2 functions, depending on the click or double-click ?
That will make tooltip and use more natural.

Nadahar:
@Sami32 It can, but how are users supposed to guess that? Most will never read tooltips...

Sami32:
Really ?

So that leave one easy solution. When a scan is run with an empty shared folder, the default ones will be set, with eventually an advertisement though i don't think it is necessary as seen the folders added should be self explanatory.

Sami32:
I also did have an other before, but as it ask to make a button color different version, i didn't considered worst the effort to make a new colored button that will offer to add default shared folder.
I mean that at first start or when empty shared folder is met, the ADD button will change color, so also if users don't read the tooltip, they will suspect something happen ;)

Nadahar:
@Sami32 I'm not sure I follow you. Changing the color of the button isn't that bad a task, but I don't understand how changing color will make users understand the logic.

The "scan" button isn't really relevant to this, there's no need to ever use it as folders will be "scanned on demand". The only reason to use the "scan" button is if you for some reason want everything to be parsed/updated immediately.

Sami32:
Not only changing the color, i proposed a tooltip but you said that users will not read it, but the color change will imply the the click on that button will ask if default shared folder should be set other way leave the user choose the ones he want.

Year, scan button use was just an other switch idea.

Nadahar:
@Sami32 My general idea is that a GUI needs to be reasonably intuitive, not a "riddle" that needs to be solved. I've been trying to twist my mind into finding a way to indicate if the shared folders table is "editable" or "standard", but I haven't been able to.

I don't think I understand your suggestion completely, could you take a screenshot and do some simple sketching or something like that? We both know that it might be that I simply don't understand what you mean...

Nadahar:
@Sami32 Now I think I'm done with this PR, unless you find something else.

Sami32:
Great work as always :+1: :smile: 
I didn't told about making an other icon though, making you extra work, as i consider that using the already existing ADD button should be suffisant, but i guess that you feeled more confortable with it ?

I still see an annoying thing. When users use the default shared folders, they cannot tune the default folders monitoring without editing.
I don't think that folders monitoring should be linked to folders editing.
What do you think ?

Nadahar:
@Sami32 I made it that way on purpose. If the configured shared folders are empty (just delete all entries and switch to "default" mode) when you switch to "manual" mode, you will be asked if you want to copy the default entries. If you answer "Yes", monitoring can be edited.

The reason for this is "technical", I keep the "manual" configuration if the default folders are used. That only goes for the monitored folders, and thus there is no where to "store" which default folders are monitored without messing with the "manual" configuration. I don't know if you understood that, but I hope so 😇 

I plan to remove the whole concept of "monitored" folders later on, when I refactor scanning and parsing. If this is implemented properly I don't think there's a reason to have this choice, every shared folder will be "monitored" - just more effectively.

Sami32:
No need to ask me to use Reviewable...

If i didn't did, it is because i don't know how to draw something when nothing visible is changed, and also because i thought that it could be an eternal explanation way that could have overflow Github without more understanding between us, so i just resigned.

Nadahar:
Reviewed 12 of 24 files at r1, 45 of 51 files at r4, 3 of 3 files at r5.

---

*[src/main/resources/i18n/messages.properties, line 436 at r1](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/26#-KykNbrV02bPuRsqUD-2-r1-436:-Kz3ubaXIVslK1OjlH8R:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/54cc941c0eadeaf20e239e5849792691ccb12d37/src/main/resources/i18n/messages.properties#L436)):*
<details><summary><i>Previously, Sami32 (Samuel) wrote…</i></summary><blockquote>

:stuck_out_tongue: 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/digitalmediaserver/digitalmediaserver/26)*

PR #41, opened 2018-05-08
 Extend FFmpeg version parsing

Nadahar:
This extends the FFmpeg version parsing to also parse all configuration options and library versions. This enables more fine-grained control over which features are available and not.

This also implementats conditional use of ```-strict experimental``` for AAC encoding using the new information.

@Sami32 This is what I've tried to explain you before regarding the version information and how it can be used to make the code more robust.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/41)

### DIFF
--- a/src/main/java/net/pms/encoders/TsMuxeRVideo.java
+++ b/src/main/java/net/pms/encoders/TsMuxeRVideo.java
@@ -30,6 +30,7 @@ import java.awt.event.ItemListener;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.util.ArrayList;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
@@ -43,6 +44,9 @@ import net.pms.configuration.DeviceConfiguration;
 import net.pms.configuration.ExecutableInfo;
 import net.pms.configuration.ExecutableInfo.ExecutableInfoBuilder;
 import net.pms.configuration.ExternalProgramInfo;
+import net.pms.configuration.FFmpegExecutableInfo;
+import net.pms.configuration.FFmpegExecutableInfo.Codec;
+import net.pms.configuration.FFmpegExecutableInfo.CoderFlags;
 import net.pms.configuration.PmsConfiguration;
 import net.pms.configuration.RendererConfiguration;
 import net.pms.dlna.*;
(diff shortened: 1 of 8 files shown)


Nadahar:
@Sami32 I would be very nice if you could take a look at this, since it's needed for #46.

src/main/java/net/pms/encoders/TsMuxeRVideo.java
```diff
@@ -465,19 +469,39 @@ public ProcessWrapper launchTranscode(
...
+							tempFFmpegCommands.add("-c:a");
+							tempFFmpegCommands.add("aac");
+							ExecutableInfo executableInfo = ffmpeg.getExecutableInfo();
+							if (executableInfo instanceof FFmpegExecutableInfo) {
+								Codec aac = ((FFmpegExecutableInfo) executableInfo).getCodecs().get("aac");
+								if (aac.containsEncoderOption(CoderOptions.EXPERIMENTAL)) {
+									tempFFmpegCommands.add("-strict");
+									tempFFmpegCommands.add("experimental");
+								}
+							}
```
Nadahar: @Sami32 Even though you don't understand all of this or even the point, if you look at the section above (from ```if (executableInfo instanceof FFmpegExecutableInfo)```) you might see the point of all this.

With the new refactoring, I can now check the properties of every codec/encoder/decoder for that FFmpeg executable (in addition to supported formats and protocols). Here I simply "get" the AAC codec, checks if its encoder is experimental and adjust the parameters accordingly.

src/main/java/net/pms/encoders/TsMuxeRVideo.java
```diff
@@ -465,19 +469,50 @@ public ProcessWrapper launchTranscode(
...
+								 *     return libavcodecVersion.isLessThanOrEqualTo(new Version(57, 16, 101, 0));
+								 * } // Libav has yet to make it non-experimental
+								 * return true;
+								 */
+								Codec aac = ((FFmpegExecutableInfo) executableInfo).getCodecs().get("aac");
+								if (aac.containsEncoderFlag(CoderFlags.EXPERIMENTAL)) {
+									tempFFmpegCommands.add("-strict");
+									tempFFmpegCommands.add("experimental");
+								}
+							}
```
Nadahar: @Sami32 I decided to include the "other way" (by looking at the library version) of determining the ```aac``` encoder experimental status as a comment. The point with this is just to remember the possibilities. When this is taken in use for various settings later the example code can be removed.

src/main/java/net/pms/encoders/TsMuxeRVideo.java
```diff
@@ -465,19 +469,50 @@ public ProcessWrapper launchTranscode(
...
+							tempFFmpegCommands.add(params.timeseek > 0 ? Double.toString(params.timeseek) : "0");
+							tempFFmpegCommands.add("-i");
+							tempFFmpegCommands.add(filename);
+							tempFFmpegCommands.add("-ac");
+							tempFFmpegCommands.add(Integer.toString(channels));
+							tempFFmpegCommands.add("-f");
+							tempFFmpegCommands.add("adts");
+							if (singleMediaAudio) {
+								tempFFmpegCommands.add("-y");
+								tempFFmpegCommands.add("-y");
```
Nadahar: @Sami32 This PR is ready except for the ```-y```s. Is it safe to just remove them, were they simply use to "fill" slots in the array?

src/main/java/net/pms/encoders/TsMuxeRVideo.java
```diff
@@ -465,19 +469,47 @@ public ProcessWrapper launchTranscode(
...
+								 *     // FFMpeg made it non-experimental in version libavcodec 57.16.101 (d9791a8656b5580756d5b7ecc315057e8cd4255e)
+								 *     return libavcodecVersion.isLessThanOrEqualTo(new Version(57, 16, 101, 0));
+								 * } // Libav has yet to make it non-experimental
+								 * return true;
+								 */
+								Codec aac = ((FFmpegExecutableInfo) executableInfo).getCodecs().get("aac");
+								if (aac.containsEncoderFlag(CoderFlags.EXPERIMENTAL)) {
+									tempFFmpegCommands.add("-strict");
+									tempFFmpegCommands.add("experimental");
+								}
```
Nadahar: @Sami32 I decided to include the "other way" (by looking at the library version) of determining the aac encoder experimental status as a comment. The point with this is just to remember the possibilities. When this is taken in use for various settings later the example code can be removed.

Nadahar:
@Sami32 I've remove the extra ```-y```s as I'm 99% sure they were just there to "fill" the array. I left the last one so there won't be any "can't overwrite" issues with the pipe.

I think I'm done with this now. Even though I know you don't see the point and don't understand everything I've done, could you please just quick look and see if you find something that you object to? If we miss something here, it's not the end of the world as only the quite unstable tsMuxeR code should be affected anyway.

Nadahar:
Reviewed 2 of 7 files at r1, 2 of 6 files at r2, 3 of 4 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/41)*

PR #42, opened 2018-05-09
Update Roku configurations

Nadahar:
This is a rebase/cleanup of https://github.com/UniversalMediaServer/UniversalMediaServer/commit/79a4c750a4f3735127179dc0b156b15f74c0365a,  https://github.com/UniversalMediaServer/UniversalMediaServer/commit/60f8de4cd37e58f544588c89dbc359503019ee20,  https://github.com/UniversalMediaServer/UniversalMediaServer/commit/041ea221e236c5d0e4747ca83c8e99421aa89cd0 and  https://github.com/UniversalMediaServer/UniversalMediaServer/commit/39e6259c1261cc642b409d7d300b2a9d0174f69b.

It looks like there still are [some problems with some of the Rokus](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=12446&p=36343&hilit=roku#p36343).

@Sami32 I assume that you know much more about this than me, and that you might already have some Roku configuration hidden somewhere, so please make corrections as you see fit.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/42)


Sami32:
Yes i have many since long, but i don't understand your push, as you have said that you didn't want add any more renderer because you plan a refactoring. Well, at least it is what i understood.
Anyway, i'm not against that but i'm on the pom.xml refactoring now so if you can wait...

Sami32:
Other issue not say here are related to our wrong DLNA profiles handling.
Here if we had log we should see, like some other renderers, some doing renderer and server, 2 different strings, depending if it is only querying ProtocolInfo phase or streaming phase.

Nadahar:
@Sami32 The renderer refactoring is a bit ahead in time, so we'll have to live when them in the meanwhile. I've just been saying that putting a lot of efforts into making renderer configurations isn't worth it, but I see no reason not to update or add trivial things.

If there are issues not caused by the configuration itself but by other bugs or missing/wrong implementations, I don't think that is related to this PR. We should only try to get the configurations as correct as possible, and then the other issues will be resolved when those things are reworked.

src/main/external-resources/nsis/setup.nsi
```diff
@@ -428,6 +428,8 @@ Section "Uninstall"
 	Delete /REBOOTOK "$INSTDIR\renderers\Realtek.conf"
 	Delete /REBOOTOK "$INSTDIR\renderers\Roku-Roku3-3.conf"
 	Delete /REBOOTOK "$INSTDIR\renderers\Roku-Roku3-5.conf"
+	Delete /REBOOTOK "$INSTDIR\renderers\Roku-Roku3-6-7.conf"
+	Delete /REBOOTOK "$INSTDIR\renderers\Roku-TV8.conf"
```
Sami32: That will conflict with the installer PR later so i don't see the point of adding it now.

src/test/java/net/pms/configuration/RendererConfigurationTest.java
```diff
@@ -129,7 +129,23 @@ public void testKnownHeaders() throws ConfigurationException, InterruptedExcepti
 
 		// FIXME: Actual conflict here! Popcorn Hour is returned...
 		//testHeaders("Realtek", "User-Agent: POSIX UPnP/1.0 Intel MicroStack/1.0.2718, RealtekMediaCenter, DLNADOC/1.50");
-		testHeaders("Realtek", "User-Agent: RealtekVOD neon/0.27.2", "RealtekMediaCenter: RealtekVOD");
+		testHeaders(
```
Sami32: I never agree on this part. For me it is just unnecessary code that give nothing more but crowd the code.
I can add just 50 lines for Roku...
Define correctly the string set in the renderers will be more efficient for now, but that should be only me...

Nadahar:
Reviewed 1 of 3 files at r1, 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/42)*

PR #47, opened 2018-05-26
Refactor database handling

Nadahar:
@Sami32 This is my "database work". There are some changes here that aren't related to the database, but that's they way it goes when making things in the reverse order :wink: Those changes are minor anyway.

Both the database initialization, verification and error handling is rewritten from scratch. So are the command line arguments and their parsing/handling. There are even some new options, and I've implemented the "standard" ```--help``` and ```-version``` arguments to make it more "command line friendly".

While testing most of the command line arguments should be fairly straight forward, testing the database error handling might be more tricky. You actually need an invalid database file to test most of the cases. If you're interested I can send you one.

It should be possible to test the "database already in use" handling simply by turning off the "one instance" restriction though.

As far as I know I'm finished with this, and any testing of reviews are appreciated.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)


Nadahar:
I see that there is a test that fails some times. I'm pretty sure it's caused by the new initialization and that it's somewhat random, as I built this without issues locally. I'll take a look at it.

Nadahar:
As I suspected there were initialization problems with the tests. I've made some fixes, but the test initialization is basically still a mess that needs refactoring in the future. It's not threadsafe and will fail randomly, although hopefully pretty seldom.

src/main/external-resources/DMS.conf
```diff
@@ -156,7 +156,7 @@ hostname =
 # ------------------------
 # Sets the port DMS will listen on; usually should not need to be set unless
 # another service has already grabbed the port.
-# Default: "", which means DMS will automatically use port 5001.
+# Default: "", which means DMS will automatically use port 5252.
```
Sami32: Why?

src/main/external-resources/DMS.conf
```diff
@@ -1147,6 +1147,21 @@ flac_path =
 #     Linux: vlc + system PATH
 vlc_path =
 
+# Database cache size
+# -------------------
+# Specifies the database cache size in MiB. This setting overrides "database
+# cache percent". If set to 0, database caching is disabled.
+# Default: ""
```
Sami32: Is that will give equivalent result as the default given below?

src/main/external-resources/DMS.conf
```diff
@@ -1248,7 +1263,7 @@ upnp_enable =
 # in case the DHCP is used in the network and the address for the PC running DMS
 # is assigned dynamically.
 # Example: 7000, which means the web interface is at localhost:7000
-# Default: 9001
+# Default: 6363
```
Sami32: Why?

src/main/external-resources/DMS.conf
Nadahar: Why not? There's nothing special about these ports, and I looked up some ports that's not actually used by anything else, according to Wikipedia. Also, it would mean that UMS and DMS wouldn't conflict if installed on the same computer.

src/main/external-resources/DMS.conf
Nadahar: Only if the database cache size isn't defined, database cache percent it used. It's not defined by default, which means that database cache percent, or its default, is applied. If you set this to any value, that cache size will be used and the percent ignored, if you set this to 0, the cache is disabled and the percent ignored.

src/main/external-resources/DMS.conf
Nadahar: Same answer as for the other port

src/main/java/net/pms/service/ProcessManager.java
```diff
@@ -126,6 +127,13 @@ public void stop() {
 		}
 	}
 
+	@Override
+	public boolean isAlive() {
+		synchronized (incoming) {
+			return terminator != null && terminator.isAlive();
```
Sami32: No he is dead, i saw it ;)

src/main/java/net/pms/util/CoverArtArchiveUtil.java
```diff
@@ -69,993 +73,1016 @@
...
+			added = true;
+		}
 
-			if (AudioUtils.tagSupportsFieldKey(tag, FieldKey.MUSICBRAINZ_ARTISTID)) {
-				artistId = tag.getFirst(FieldKey.MUSICBRAINZ_ARTISTID);
+		if (isNotBlank(tagInfo.artistId)) {
+			if (added) {
+				query.append(and);
+			}
+			query.append("arid:").append(tagInfo.artistId);
```
Sami32: `artistid` will have be less confusing IMO.

src/main/java/net/pms/util/TitleCase.java
```diff
@@ -0,0 +1,227 @@
...
+		@Nullable Collection<? extends CharSequence> keepLower
+	) {
+		return convert(s, false, locale, keepLower);
+	}
+
+	/**
+	 * Converts the specified {@link String} to title-case.
+	 *
+	 * @param s the {@link String} to convert.
+	 * @param stripNewLines if {@code true} newlines are replaces with spaces.
```
Sami32: `replaced` instead of `replaces`?

src/main/java/net/pms/newgui/DatabaseProblem.java
```diff
@@ -0,0 +1,572 @@
...
+	 */
+	public DatabaseProblem(@Nullable Component parentComponent, @Nonnull TableManager tableManager) {
+		if (PMS.isHeadless()) {
+			throw new IllegalStateException("DatabaseProblem dialog can't run in headless mode");
+		}
+		this.parentComponent = parentComponent;
+		this.tableManager = tableManager;
+		this.locale = PMS.getLocale();
+
+		// Apply the orientation for the locale
```
Sami32: What do that mean?

src/main/java/net/pms/Messages.java
```diff
@@ -126,41 +144,42 @@ public static String getString(String key, Locale locale) {
...
 
 	/**
-	 * Checks if the given <code>Locale</code> should use the root language
-	 * file (messages.properties) which is en-US. Currently that is all variants
-	 * of English but British English.
-	 * @param locale the <code>Locale</code> to check
-	 * @return The result
+	 * Checks if the given {@link Locale} should use the root language file (
+	 * {@code messages.properties}) which is {@code en-US}. Currently that is
+	 * all variants of English but British English.
```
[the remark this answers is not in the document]
Nadahar: @Sami32 When "but" is used that way, it means "anything else than". So, it says "all variants of English except British English".

src/main/java/net/pms/PMS.java
```diff
@@ -1009,91 +1206,259 @@ public Format getAssociatedFormat(String filename) {
...
+			if (!argument.startsWith("-")) {
+				argument = argument.toLowerCase(Locale.ROOT);
+			}
+			switch (argument) {
+				case "-h":
+				case "--help":
+				case "help":
+					result.put(Option.HELP, null);
+					break;
+				case "-V":
```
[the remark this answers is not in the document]
Nadahar: @Sami32 Because ```-v``` is in use and has become a de factor standard for "verbose" at least in Linux. Since the command line options are most relevant under Linux, I chose to respect that practice.

src/main/java/net/pms/PMS.java
```diff
@@ -1009,91 +1206,259 @@ public Format getAssociatedFormat(String filename) {
...
+				case "-c":
+				case "--console":
+				case "--headless":
+				case "headless":
+				case CONSOLE:
+					result.put(Option.HEADLESS, null);
+					break;
+				case "-n":
+				case "--nativelook":
+				case NATIVELOOK:
```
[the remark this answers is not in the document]
Nadahar: I have no idea what this actually does, it's ancient stuff from PMS that I haven't investigated. It might be that it does nothing today, or it might be that it can give a better GUI under some circumstances. I've just kept it as it was, only adapted the option to the new "style".

src/main/java/net/pms/PMS.java
```diff
@@ -1009,91 +1206,259 @@ public Format getAssociatedFormat(String filename) {
...
+				case "--scrollbars":
+				case SCROLLBARS:
+					result.put(Option.SCROLLBARS, null);
+					break;
+				case "-C":
+				case "--noconsole":
+				case NOCONSOLE:
+					result.put(Option.NOCONSOLE, null);
+					break;
+				case "-v":
```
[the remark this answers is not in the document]
Nadahar: Because ```-v``` has become a de factor standard for "verbose" or "extra logging", at least in Linux. As you can see, ```-t```, "--trace``` and ```trace``` is supported as well.

src/main/java/net/pms/database/TableCoverArtArchive.java
```diff
@@ -145,25 +197,25 @@ public static void writeMBID(final String mBID, final byte[] cover) {
 	}
 
 	/**
-	 * Looks up cover in the table based on the given MBID. Never returns
-	 * <code>null</code>
+	 * Looks up cover in the table based on the given MBID and returns the
```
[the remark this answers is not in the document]
Nadahar: It depends on where it is. ```MBID``` is used as an acronym/initialism for "MusicBrainz Identifier". If it's used in a comment or JavaDocs, this form is used. However, when it's used as a variable/field/method name, Java convention dictates that lower camel case should be used. That means that it can't start with a capital letter. I've seen many different ways to "solve" the combination of acronyms and lower camel case, and I don't even think always I manage to be consistent myself, by my idea is that it represents the initials from ```musicBrainzIDentifier```. When I think about it, the ```D``` should be lower case in all uses of ```ID```, so ```Id``` is more correct IMO. But, for some reason "identification" or "identifier" is commonly abbreviated [ID](https://en.wikipedia.org/wiki/Identifier), so I guess I'll have to respect that :smile:

src/main/java/net/pms/PMS.java
```diff
@@ -1009,91 +1206,259 @@ public Format getAssociatedFormat(String filename) {
...
+				case "--nativelook":
+				case NATIVELOOK:
+					result.put(Option.NATIVE_LOOK, null);
+					break;
+				case "-s":
+				case "--scrollbars":
+				case SCROLLBARS:
+					result.put(Option.SCROLLBARS, null);
+					break;
+				case "-C":
```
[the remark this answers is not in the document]
Nadahar: ```--noconsole``` is kept since it's already there from PMS. My logic when picking ```-C``` was simply that ```-n``` wasn't available, and that a capital letter is often (but far from always) use as "not", so logically it would represent ```--!console```. ```-g```/```-GUI``` might be used as well, but it's not exactly the same. This isn't an option that enables the GUI, it's an option that makes DMS fail to start if no GUI can be created. "Force GUI" would perhaps fit better. I'm open to suggestions as long as we keep ```--noconsole``` and ```noconsole``` for backwards compatibility.

src/main/java/net/pms/database/TableCoverArtArchive.java
```diff
@@ -58,51 +54,105 @@
...
+			connection.rollback();
+			throw e;
+		} finally {
+			connection.setAutoCommit(true);
+		}
 	}
 
 	/**
-	 * Stores the cover {@link Blob} with the given mBID in the database
+	 * Stores the cover {@link Blob} with the given mBID in the database.
```
Sami32: As i said:
`mBID`

src/main/java/net/pms/database/TableCoverArtArchive.java
```diff
@@ -58,51 +54,105 @@
...
 	 *
 	 * @param mBID the MBID to store
 	 * @param cover the cover as a {@link Blob}
+	 * @throws IllegalArgumentException If {@code mBID} is blank.
 	 */
-	public static void writeMBID(final String mBID, final byte[] cover) {
+	public void writeMBID(@Nonnull String mBID, @Nullable byte[] cover) {
+		if (isBlank(mBID)) {
+			throw new IllegalArgumentException("mBID cannot be blank");
```
Sami32: `mBID`

src/main/java/net/pms/database/TableId.java
```diff
@@ -0,0 +1,155 @@
...
+	 */
+	@Nonnull
+	public String getName() {
+		return name;
+	}
+
+	/**
+	 * Gets the display name for the {@link Table} represented by this
+	 * {@link TableId}. This is either a "prettified" version of
+	 * {@link #getName()} or a "root"/English name depending on if the
```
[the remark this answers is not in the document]
Nadahar: No, although it's kind of related. It means the "root resource bundle", which is a Java thing. The translations are loaded as a resource bundle with support for different locales, and there's one "root" bundle that is the fallback if no more suitable bundle can be found. This is the representation of the file ```messages.properties``` in our case, which is the "standard/US English" version. We do have ```messages_en-GB.properties``` that only defines the strings that is different from the "root"/US English strings. It simply means that this method will always get the strings from ```messages.properties``` without taking locale into consideration, and is typically used for logging (which we don't want to be localized).

src/main/java/net/pms/newgui/ErrorDialog.java
```diff
@@ -0,0 +1,301 @@
...
+		dialog.dispose();
+
+		if (!Integer.valueOf(JOptionPane.OK_OPTION).equals(pane.getValue())) {
+			aborted = true;
+		}
+	}
+
+	private JComponent buildComponent() {
+		// UIManager manages to get the background color wrong for text
+		// components on OS X, so we apply the color manually
```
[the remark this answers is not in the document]
Nadahar: It's sad to see what Oracle's goal is, but at least it explains why they've fucked up Java with Java 9 and 10. They've simply denied access to stuff they intend to make "licensed customers only".

I don't quite see how this relates to the code in question though.

src/main/java/net/pms/util/TitleCase.java
Nadahar: Yes... my fingers constantly does things like this on their own :blush:

src/main/java/net/pms/database/TableManager.java
```diff
@@ -0,0 +1,556 @@
...
+			LOGGER.error("Database tables check failed with: {}", e.getMessage());
+			LOGGER.trace("", e);
+			lastException = e;
+			clearConnectionPool();
+		}
+	}
+
+	/**
+	 * Stops this {@link TableManager}. This will cause the connection pool to
+	 * shut down and this {@link TableManager} to go into a disconnected state.
```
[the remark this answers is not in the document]
Nadahar: Yes, I think "shutdown" would be wrong here.

src/main/java/net/pms/database/TableManager.java
```diff
@@ -0,0 +1,556 @@
...
+			databaseFile = new File(databaseName);
+		}
+		return databaseFile.getAbsolutePath();
+	}
+
+	private static String buildURL(@Nonnull String databaseFilename, @Nullable PmsConfiguration configuration) {
+		StringBuilder sb = new StringBuilder(Constants.START_URL)
+			.append("nio:")
+			.append(databaseFilename)
+			.append(";MULTI_THREADED=1");
```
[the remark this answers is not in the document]
Nadahar: Hmm, I'm a bit unsure about this. Our use will be lots of reads and a few writes, so I'm not sure the situation in the above test case applies. It's still bad if this leads to potential corruption of the database, but it's very tempting to try and see how it goes.

src/main/java/net/pms/util/CoverArtArchiveUtil.java
Nadahar: It's the standard lower camel case, as customary for Java. Written out, it would be two words "artist ID", and thus it will have to become ```artistID``` or ```artistId```. Don't ask me why I chose the latter here :stuck_out_tongue:

src/main/java/net/pms/newgui/DatabaseProblem.java
Nadahar: It means to set the RTL or LTR for the dialog.

src/main/java/net/pms/newgui/ErrorDialog.java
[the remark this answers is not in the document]
Sami32: Hehe...i was just sharing informations.
Seing the UIManager make me remember the Swing issue with Java 9 and 10, as Oracle removed some libraries support, so after september we shouldn't have anymore bad surprise, except if Oracle do sabotage with Java 8...

src/main/java/net/pms/database/TableCoverArtArchive.java
Nadahar: That on is correct as it references the variable, not the acronym.

src/main/java/net/pms/database/TableCoverArtArchive.java
Nadahar: I agree that this should be changed, although I probably referred to the parameter when I wrote this.

src/main/java/net/pms/newgui/DatabaseProblem.java
Sami32: Ahhh OK.
Then "direction" word should have be choose.
https://www.w3.org/International/questions/qa-scripts
https://en.wikipedia.org/wiki/Right-to-left

src/main/java/net/pms/database/TableCoverArtArchive.java
Sami32: I admit that i was not sure about that one.

src/main/java/net/pms/newgui/DatabaseProblem.java
Nadahar: Well, this is just code I have copied from ```LooksFrame```, but I think it's natural to use the same term as Java uses. I think your complaint should be directed at [Oracle](https://docs.oracle.com/javase/7/docs/api/java/awt/ComponentOrientation.html).

src/main/java/net/pms/database/TableCoverArtArchive.java
Nadahar: @Sami32 GitHub is ratting us out. They have now made all edits public :stuck_out_tongue:

Sami32: Lol. Good luck to them with me then ;)

src/main/java/net/pms/newgui/DatabaseProblem.java
Nadahar: I agree, but I must say that in this instance I'm with Oracle (don't quote me on that :stuck_out_tongue_closed_eyes:). I think "orientation" is a better fit than "direction", and I suspect that something is "lost in translation" to French :wink:

src/main/java/net/pms/util/CoverArtArchiveUtil.java
Sami32: You misunderstood me. I was talking about this:
"arid:"

src/main/java/net/pms/newgui/DatabaseProblem.java
Sami32: I just meant that it would be nice if Oracle could stick to standards, whatever you agree or not with it 😉

src/main/java/net/pms/database/TableCoverArtArchive.java
```diff
@@ -100,8 +100,8 @@ protected void createTable(@Nonnull Connection connection) throws SQLException {
 	 * The changes for each version in the form of {@code ALTER TABLE} must be
 	 * implemented here.
 	 * <p>
-	 * For implementation example see
-	 * {@link TableMusicBrainzReleases#upgradeTable(Connection, int)}
+	 * For an implementation example see
```
Sami32: Oups...i've missed that one. Too late review i guess ;)

src/main/java/net/pms/database/TableCoverArtArchive.java
```diff
@@ -133,10 +133,10 @@ protected void upgradeTable(@Nonnull Connection connection, int currentVersion)
 	}
 
 	/**
-	 * Stores the cover {@link Blob} with the given mBID in the database.
+	 * Stores the cover {@link Blob} with the given {@code MBID} in the database.
```
Sami32: Ok, now you loose me. Shouldn't have been `mBID`?

src/main/java/net/pms/util/CoverArtArchiveUtil.java
Nadahar: Ah, then you need to complain to MusicBrainz.org to get them to change their API: https://musicbrainz.org/doc/Indexed_Search_Syntax#Query_syntax

src/main/java/net/pms/database/TableCoverArtArchive.java
Nadahar: Lol, you must make up your mind :wink:  I just changed it because you complained that I referred to the parameter, not the acronym. I added ```{@code }``` simply because I had forgotten, I use that on both variable references and acronyms/specific names/terms.

Sami32: > I added {@code } simply because I had forgotten, I use that on both variable references and acronyms/specific names/terms.

I didn't knew so this add confused me.

src/main/java/net/pms/util/CoverArtArchiveUtil.java
Sami32: You are right. What a disaster choice. That said i like complain ;P

Sami32:
Probably not related to your PR but i still get the following exception in the`DMS.trace.db`:
```
2018-06-02 16:02:58 jdbc[3]: exception
org.h2.jdbc.JdbcSQLException: Table "FILES" not found; SQL statement:
SELECT count(*) FROM FILES [42102-196]
2018-06-02 16:02:58 jdbc[3]: exception
org.h2.jdbc.JdbcSQLException: Table "FILES" not found; SQL statement:
DROP TABLE FILES [42102-196]
```

Sami32:
**Cancel** button is preselected at the end, but i think that **Continue** should be the one preselected, as it is already the 3th step and enter could be used.

Nadahar:
> i still get the following exception in theDMS.trace.db while using your test files:

That should be related to this PR. What did you do to get these? Did you first downgrade the database and then started up, or what happened?

Nadahar:
> Cancel button is preselected at the end, but i think that Continue should be the one preselected, as it is already the 3th step and Enter key could be used.

That should be fixed in 75d5c483edab97160e5ee1940c737c53366072ec.

Nadahar:
@Sami32 Regarding the exception, are you getting this exception only in the "trace" file? Not everything there is an error, I think ```count(*)``` is used to simply check if the table is there or should be created during initialization of ```DLNAMediaDatabase```. If the table doesn't exist this exception would be caught and the table created.

Sami32:
Yes, it was after having dowgraded the database, and only in the _DMS.trace.db_ file.

The **information** icon displayed at the downgrading time should be replaced by the **warning** icon.

Nadahar:
I didn't know there was an information icon until after it was solved. The dialog has an error icon by default, as AFAIK it only changes when the database is ok.

Sami32:
capture_06022018_225330

Nadahar:
@Sami32 Yes, that's not with the latest code, since it should now be a warning icon. But, it was a question mark also on XP :wink:

Nadahar:
@Sami32 If completed my updates. It would be nice if you took a quick look at the changes in case there's something I overlooked, but most of the code is unchanged. Almost all my changes are found in 91d4f4b9fbbe9485b0863a6ef68b6b99ccfac2d0, 88c192e738f4cf7bf0e7f681e171e5b899a871a9 and ecf794396b973feb9583b14855e4ab435e7917c3.

Nadahar:
Pun intended? :innocent:

Sami32:
😇

Nadahar:
Reviewed 92 of 101 files at r1, 1 of 4 files at r2, 15 of 15 files at r3.
Dismissed @Sami32 from 12 discussions.
Review status: all files reviewed, 9 unresolved discussions (waiting on @Nadahar and @Sami32)

---

*[src/main/java/net/pms/database/TableCoverArtArchive.java, line 136 at r1](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47#-LDTeUB-dvNo4PRntMUM-r1-136:-LElVzMp4I5d8Wye_2F6:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/0f223a5b05c354b7e24dc443197e4c5b04463421/src/main/java/net/pms/database/TableCoverArtArchive.java#L136)):*

I agree that this should be changed, although I probably referred to the parameter when I wrote this.
</blockquote></details>

Done.

---

*[src/main/java/net/pms/util/TitleCase.java, line 101 at r1](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47#-LDTeUB-dvNo4PRntMVi-r1-101:-LElacl21Mo12MuZoF9R:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/0f223a5b05c354b7e24dc443197e4c5b04463421/src/main/java/net/pms/util/TitleCase.java#L101)):*

Yes... my fingers constantly does things like this on their own :blush: 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)*

Nadahar:
Dismissed @Sami32 from 2 discussions.
Review status: all files reviewed, 7 unresolved discussions (waiting on @Sami32)

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)*

Nadahar:
Dismissed @Sami32 from a discussion.
Review status: all files reviewed, 6 unresolved discussions (waiting on @Sami32)

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)*

Nadahar:
Dismissed @Sami32 from a discussion.
Review status: all files reviewed, 5 unresolved discussions (waiting on @Sami32)

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)*

Nadahar:
Dismissed @Sami32 from 4 discussions.
Review status: all files reviewed, 1 unresolved discussion (waiting on @Sami32)

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)*

Nadahar:
Dismissed @Sami32 from a discussion.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/47)*

PR #52, opened 2018-07-14
DSD

Sami32:


Sami32:
Note! 
As this new DSD code is only for more recent MediaInfo versions this PR shouldn't be merged now without adding old version supportas before.

Sami32:
That is said, it is not mandatory to support DSD for older MI versions but Linux users not using a flat package could possibly miss this support.
I pass the hand for this job.

PR #53, opened 2018-07-15
Atmos, TrueHD and E-AC-3 fixes

Sami32:
Make our code ready for the following MediaInfo versions.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/53)


Sami32:
Could not be need if it get fixed before the next release, so better wait some few days.
See https://github.com/MediaArea/MediaInfoLib/issues/923#issuecomment-405106807

JeromeMartinez:
It is a moving part, I think the safest way for you is to check as you do + check start with same string in profile part (for hybrid content). So you'll be sure whatever is the final choice in MediaInfo.

JeromeMartinez:
Atmos: safer to use "contains".

Sami32:
Ok i see, a really moving part then 😉

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -706,7 +706,9 @@ private static void getFormat(
 			format = FormatConfiguration.TRUEHD;
 		} else if (value.equals("e-ac-3")) {
 			format = FormatConfiguration.EAC3;
-		} else if (value.contains("atmos") || value.contains("e-ac-3 joc") || value.equals("131")) {
+		} else if (value.contains("mlp")) {
```
Sami32: @JeromeMartinez Thank you very much for your explaination as it wasn't clear to me :+1:

PR #55, opened 2018-07-16
Disable Chromecast by default

Sami32:
As it give more issue than benefice it should only be enable by users having a need for it.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/55)


Nadahar:
I agree, but shoudn't the documentation be updated as well?

PR #60, opened 2018-07-25
Fix OGA in web renderer

### LINKED ISSUE #62: OGG files without video stream are considered as video
The OGG files without video stream have to be considered as OGA in the web renderer.
Because as it is now they fail to play.

Sami32:
Need a Java expert to fix the OGG files without video stream to be considered as OGA in the web renderer.

😉

This should now fix #62

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/60)

### DIFF
--- a/src/main/java/net/pms/configuration/WebRender.java
+++ b/src/main/java/net/pms/configuration/WebRender.java
@@ -39,6 +39,7 @@ import net.pms.encoders.Player;
 import net.pms.external.StartStopListenerDelegate;
 import net.pms.formats.*;
 import net.pms.formats.audio.MP3;
+import net.pms.formats.audio.OGA;
 import net.pms.formats.image.BMP;
 import net.pms.formats.image.GIF;
 import net.pms.formats.image.JPG;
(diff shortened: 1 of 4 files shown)


Nadahar:
Damn, where should we find one of those?

Nadahar:
This doesn't seem to be a small fix, since the web interface completely ignores the normal parsing and is only based on file extensions. If must somehow be refactored so that it uses the normal parsing, where you have access to the `DLNAMediaInfo` instance where this has already been resolved.

Sami32:
Ok, so as it doesn't need to be added in this PR i will post a specific issue for that.

Sami32:
Thank you for having looked at it 👍

Nadahar:
@Sami32 What will this PR do without it? Only work for files with an `OGA` extension?

Sami32:
Yes

Sami32:
Yes, but it is all i can fix by myself by now.

Nadahar:
This should also fix #62

Nadahar:
The "requested changes" are simply testing.

Sami32:
Hehe, work great 👍 

I knew that we have a Java expert around 😉

PR #67, opened 2018-08-20
Attempt to enhance systray icon display issue

Sami32:
Related to that issue report:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/965

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/67)


Sami32:
No the correct way to handle it.

Nadahar:
@Sami32 Please reopen and reuse this PR if you make a new attempt. You can just force push any branch over this branch, *as long as the PR is open*.

Sami32:
Worked fine.
Though no tests has been done for HiDPI.

src/main/java/net/pms/io/BasicSystemUtils.java
```diff
@@ -279,7 +279,7 @@ protected Image resolveTrayIcon() {
 	 */
 	@Nonnull
 	protected String getTrayIconName() {
-		return "icon-24st.png";
+		return "icon-24.png";
```
Sami32: I suspect that you didn't respected my icon so it will not look nice, native on all the Linux desktop system tray...

Sami32: 2px up and right and 1px down and left if i remember correctly.

Nadahar: I've finally pushed a commit so all the icons are updated. I did not replicate that it wasn't centered, as I thought that was just a "bug" in your image.

I find it strange that Linux requires every icon to be off center. It's easy to do, but do you have any references for that? I prefer to have just one icon for each size if possible and avoid "special versions".

I have done 4px on the largest, then 2px from 256px down to >32px, 1px from 32 to >16px and 0px for 16 and 16x12.

All these margins/insets are equal on all sides so that the image is still centered.

Nadahar:
@Sami32 I'm done updating the icons, so it would be nice if you could check them out as they are now and tell me what you're unhappy with.

src/main/java/net/pms/io/WindowsSystemUtils.java
```diff
@@ -358,8 +359,24 @@ public String parsePingLine(String line) {
...
+		int dpi = Toolkit.getDefaultToolkit().getScreenResolution();
+		switch (dpi) {
+			case 96:
+				return "icon-16.png";
+			case 120:
+				return "icon-20.png";
+			case 144:
+				return "icon-24.png";
+			case 168:
+				return "icon-28.png";
```
Sami32: As i never see any Microsoft documentation about that one, did Windows 10 175% test confirm that it improved the aspect over 24 or 32px notification tray icon?

Nadahar: No, but it's the logical thing to do. In fact, nothing I do helps. I've even been able to force Windows 10 into 200% (4K) now, and the icon is still scaled. I think the Java SystemTray thing can't handle DPI.

Nadahar: Ok, this is definitely a problem with Java SystemTray. Now I turned off automatic scaling in SystemTray and used the 32px icon. What happens then, is that it crops the image so that only a quarter of the image is shown (so, still 16px), and then this is enlarged and blurred :wink: 

There's nothing we can do to get this working with this SystemTray implementation.

src/main/java/net/pms/io/WindowsSystemUtils.java
```diff
@@ -358,8 +359,24 @@ public String parsePingLine(String line) {
...
+			case 192:
+				return "icon-32.png";
+			default:
+				// This will be scaled, so use a large one for a better end result
+				return "icon-256.png";
```
Sami32: It seem overskilled to me.
Windows have a theorical maximum of 480 DPI (500%) that mean 80px maximum, and as they scaling algorythm seem bad it should better to lowering it to that value or 64px or 96px because we already have them.

Nadahar: It's not only that, from what I understand it's possible to "calibrate" your DPI by giving the actual size of your monitor (if you need 1cm on the screen to be 1cm in the real world). If that is done, you can get non-standard DPI's, like 102 or 68. That's where the scaling from 256 kicks in.

Nadahar:
I consider this finished now. I've squashed my tweaks into you commit.

Nadahar:
Feel free to merge.

Sami32:
>https://docs.oracle.com/javase/9/docs/api/index.html?java/awt/SystemTray.html

    getTrayIconSize

    public Dimension getTrayIconSize​()

    Returns the size, in pixels, of the space that a tray icon will occupy in the system tray. Developers may use this methods to acquire the preferred size for the image property of a tray icon before it is created. For convenience, there is a similar method TrayIcon.getSize() in the TrayIcon class.

    Returns:
        the default size of a tray icon, in pixels
    See Also:
        TrayIcon.setImageAutoSize(boolean), Image, TrayIcon.getSize()

But it seem that it only get fixed in Java 9 onward:
https://bugs.openjdk.java.net/browse/JDK-8150176?attachmentViewMode=list

Nadahar:
Yeah, they probably won't do anything more with Java 8 - but at least that means that this code will probably start working in the future :wink:

PR #89, opened 2018-09-22
MIME types little refactoring

Sami32:
---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/89)

### DIFF
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -513,6 +513,8 @@ public class LibMediaInfoParser {
 			format = FormatConfiguration.FLV;
 		} else if (value.equals("webm")) {
 			format = FormatConfiguration.WEBM;
+		} else if (value.equals("mxf")) {
+			format = FormatConfiguration.MXF;
 		} else if (value.equals("qt") || value.equals("quicktime")) {
 			format = FormatConfiguration.MOV;
 		} else if (
@@ -605,6 +607,8 @@ public class LibMediaInfoParser {
 			format = FormatConfiguration.VP8;
 		} else if (value.startsWith("vp9")) {
 			format = FormatConfiguration.VP9;
+		} else if (value.startsWith("av1")) {
+			format = FormatConfiguration.AV1;
 		} else if (
 			(
 				value.startsWith("div") ||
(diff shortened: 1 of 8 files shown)


Sami32:
I can remove all the images MIME types if you can confirm me that these code lines will not be need for a later UPnP implementation.

Sami32:
If you meant MPEGTS, no, only a bad hardcoded hack for Sony AFAIK.

Sami32:
`dnxhd` iss used by the old MI versions:
https://github.com/MediaArea/MediaInfoLib/commit/2fe6120c9c77b8c872477a843f703b1b59630d14

Sami32:
I can see it in the commit under the draft, weird.

Nadahar:
I've added it when I resolved conflicts between my local copy of the branch and the remote one. That means that it was there earlier. My question was why you removed it.

Sami32:
No idea, it is too far now

Sami32:
No i don't see any wrong could happen.
When no container could be determined the audio codec is applied.
When the mime test after the container cases switches is null then
```
if (mimeType == null) {
			mimeType = HTTPResource.getDefaultMimeType(getMediaType());
}
```
So in any cas a default MIME Type will be applied, as a container, only if FFmpeg or MI can identify it of course.

Nadahar:
> When no container could be determined the audio codec is applied.

Where is this code? For me it seems like you removed the code that did this. Did you move it somewhere else?

Sami32:
I saw it line 1656

Nadahar:
@Sami32 You've got to link to the code you mean. If you mean `HTTPResource.getDefaultMimeType()`, that doesn't consider the codec at all. It only looks at the "type" (audio, video or image) and returns a "generic" MIME type for that "type". The only thing you can be certain of with this method, is that the returned MIME type is wrong.

Sami32:
I was not able to find any new issue in the FFmpeg parsing in my 4 tests.

But keep it in standby i'll do some more tests later.

Sami32:
I wasn't able to find any issues until now.

Nadahar:
Sounds good :+1: Let me do some testing when I can catch up with myself :innocent:

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -655,6 +661,10 @@ private static void getFormat(
 			value.equals("wmva")
 		) {
 			format = FormatConfiguration.VC1;
+		} else if (value.equals("vc-3")) {
```
Nadahar:
```suggestion
		} else if (value.equals("vc-3") || value.startsWith("dnxhd")) {
```

PR #93, opened 2018-09-29
Remove unecessary transcoding code

Sami32:


Sami32:
Note:
All the code related to `TranscodeAudioTo441kHz` should also be removed later when the MIME Types PR will be merged, as all these sync / pitch issues that this workaround is used for are only related to incorrect MIME Types and incorrect DLNA flag.

Sami32:
No wonder why:
>  http-get:*:audio/L16;rate=44100;channels=1:DLNA.ORG_PN=LPCM
  http-get:*:audio/L16;rate=44100;channels=2:DLNA.ORG_PN=LPCM

Sami32:
The 3 others poor renderers still using this parameter are not using MediaInfo that explain why they added this option from old time.
When they will be enhanced and use MediaInfo parsing the `TranscodeAudioTo441kHz` code could be totally removed.

Sami32:
Or like the `http-get:*:audio/L16:DLNA.ORG_PN=LPCM_low` DLNA profile

PR #97, opened 2018-10-03
fix_potential_errors_of_video_and_audio_streams_ffmpegvideo

onon765trb:
Fix  potential_errors

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -319,6 +319,7 @@ public class FFMpegVideo extends Player {
 		List<String> transcodeOptions = new ArrayList<>();
 		final String filename = dlna.getFileName();
 		final RendererConfiguration renderer = params.mediaRenderer;
+
 		String customFFmpegOptions = renderer.getCustomFFmpegOptions();
 
 		if (
@@ -366,13 +367,15 @@ public class FFMpegVideo extends Player {
 					transcodeOptions.add("-an");
 				} else if (type() == FormatType.AUDIO) {
 					// Skip
-				} else if (renderer.isTranscodeToAAC()) {
-					transcodeOptions.add("-c:a");
-					transcodeOptions.add("aac");
 				} else {
-					if (!customFFmpegOptions.contains("-c:a ")) {
-						transcodeOptions.add("-c:a");
-						transcodeOptions.add("ac3");
+					if (!customFFmpegOptions.matches(".*-(c:a|codec:a|acodec) [^ ]+.*")) {
+						if (renderer.isTranscodeToAAC()) {
+							transcodeOptions.add("-c:a");
+							transcodeOptions.add("aac");
+						} else {
+								transcodeOptions.add("-c:a");
+								transcodeOptions.add("ac3");
+						}
 					}
 				}
 			}
@@ -386,17 +389,27 @@ public class FFMpegVideo extends Player {
 
 			// Output video codec
 			if (renderer.isTranscodeToH264() || renderer.isTranscodeToH265()) {
-				if (!customFFmpegOptions.contains("-c:v")) {
+				if (!customFFmpegOptions.matches(".*-(c:v|codec:v|vcodec) [^ ]+.*")) {
 					transcodeOptions.add("-c:v");
 					if (renderer.isTranscodeToH264()) {
 						transcodeOptions.add("libx264");
 					} else {
 						transcodeOptions.add("libx265");
 					}
-					transcodeOptions.add("-tune");
-					transcodeOptions.add("zerolatency");
+
+					if (!customFFmpegOptions.contains("-tune")) {
+						transcodeOptions.add("-tune");
+						transcodeOptions.add("zerolatency");
+					}
+
+					//Consider encoders which have "preset" option
+					if (!customFFmpegOptions.contains("-preset")) {
+						transcodeOptions.add("-preset");
+						transcodeOptions.add("ultrafast");
+					}
+
 				}
-				if (!customFFmpegOptions.contains("-preset")) {
+				if (!customFFmpegOptions.contains("-preset") && customFFmpegOptions.matches(".*-(libx264|libx265).*")) {
 					transcodeOptions.add("-preset");
 					transcodeOptions.add("ultrafast");
 				}
@@ -404,11 +417,15 @@ public class FFMpegVideo extends Player {
 					transcodeOptions.add("-level");
 					transcodeOptions.add("31");
 				}
-				transcodeOptions.add("-pix_fmt");
-				transcodeOptions.add("yuv420p");
+				if (!customFFmpegOptions.contains("-pix_fmt")) {
+					transcodeOptions.add("-pix_fmt");
+					transcodeOptions.add("yuv420p");
+				}
 			} else if (!dtsRemux) {
-				transcodeOptions.add("-c:v");
-				transcodeOptions.add("mpeg2video");
+				if (!customFFmpegOptions.matches(".*-(c:v|codec:v|vcodec) [^ ]+.*")) {
+					transcodeOptions.add("-c:v");
+					transcodeOptions.add("mpeg2video");
+				}
 			}
 
 			if (!customFFmpegOptions.contains("-f")) {
@@ -581,34 +598,37 @@ public class FFMpegVideo extends Player {
 			videoBitrateOptions.addAll(new ArrayList<>(Arrays.asList(customOptions)));
 		} else {
 			// Add x264 quality settings
-			String x264CRF = configuration.getx264ConstantRateFactor();
-
-			// Remove comment from the value
-			if (x264CRF.contains("/*")) {
-				x264CRF = x264CRF.substring(x264CRF.indexOf("/*"));
-			}
-
-			if (x264CRF.contains("Automatic")) {
-				if (x264CRF.contains("Wireless") || maximumBitrate < 70) {
-					x264CRF = "19";
-					// Lower quality for 720p+ content
-					if (media.getWidth() > 1280) {
-						x264CRF = "23";
-					} else if (media.getWidth() > 720) {
-						x264CRF = "22";
-					}
-				} else {
-					x264CRF = "16";
-
-					// Lower quality for 720p+ content
-					if (media.getWidth() > 720) {
+			String customFFmpegOptions = params.mediaRenderer.getCustomFFmpegOptions();
+			if (!customFFmpegOptions.matches(".*-(crf|b|b:v|q|q:v|(x264opts|x264-params) +([^ =]+=)*(crf|bitrate|B|q|qp)=).*")) {
+				String x264CRF = configuration.getx264ConstantRateFactor();
+	
+				// Remove comment from the value
+				if (x264CRF.contains("/*")) {
+					x264CRF = x264CRF.substring(x264CRF.indexOf("/*"));
+				}
+	
+				if (x264CRF.contains("Automatic")) {
+					if (x264CRF.contains("Wireless") || maximumBitrate < 70) {
 						x264CRF = "19";
+						// Lower quality for 720p+ content
+						if (media.getWidth() > 1280) {
+							x264CRF = "23";
+						} else if (media.getWidth() > 720) {
+							x264CRF = "22";
+						}
+					} else {
+						x264CRF = "16";
+	
+						// Lower quality for 720p+ content
+						if (media.getWidth() > 720) {
+							x264CRF = "19";
+						}
 					}
 				}
-			}
-			if (isNotBlank(x264CRF) && !params.mediaRenderer.nox264()) {
-				videoBitrateOptions.add("-crf");
-				videoBitrateOptions.add(x264CRF);
+				if (isNotBlank(x264CRF) && !params.mediaRenderer.nox264()) {
+					videoBitrateOptions.add("-crf");
+					videoBitrateOptions.add(x264CRF);
+				}
 			}
 		}
 
@@ -1032,7 +1052,7 @@ public class FFMpegVideo extends Player {
 
 			String customFFmpegOptions = renderer.getCustomFFmpegOptions();
 
-			// Audio bitrate
+			// Audio channel bitrate sample rate
 			if (!ac3Remux && !dtsRemux && !(type() == FormatType.AUDIO)) {
 				int channels = 0;
 				if (
@@ -1055,7 +1075,7 @@ public class FFMpegVideo extends Player {
 					cmdList.add(String.valueOf(channels));
 				}
 
-				if (!customFFmpegOptions.contains("-ab ")) {
+				if (!customFFmpegOptions.matches(".*-(ab|b:a|q:a).*")) {
 					cmdList.add("-ab");
 					if (renderer.isTranscodeToAAC()) {
 						cmdList.add(Math.min(configuration.getAudioBitrate(), 320) + "k");
@@ -1064,7 +1084,7 @@ public class FFMpegVideo extends Player {
 					}
 				}
 
-				if (!customFFmpegOptions.contains("-ar ")) {
+				if (!customFFmpegOptions.contains("-ar ") && params.aid.getSampleFrequency() != params.mediaRenderer.getTranscodedVideoAudioSampleRate()) {
 					cmdList.add("-ar");
 					cmdList.add("" + params.mediaRenderer.getTranscodedVideoAudioSampleRate());
 				}


src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -581,34 +598,37 @@
...
-					} else if (media.getWidth() > 720) {
-						x264CRF = "22";
-					}
-				} else {
-					x264CRF = "16";
-
-					// Lower quality for 720p+ content
-					if (media.getWidth() > 720) {
+			String customFFmpegOptions = params.mediaRenderer.getCustomFFmpegOptions();
+			if (!customFFmpegOptions.matches(".*-(crf|b|b:v|q|q:v|(x264opts|x264-params) +([^ =]+=)*(crf|bitrate|B|q|qp)=).*")) {
```
Sami32: adding qscale and qscale:v

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -1032,7 +1052,7 @@ public synchronized ProcessWrapper launchTranscode(
 
 			String customFFmpegOptions = renderer.getCustomFFmpegOptions();
 
-			// Audio bitrate
+			// Audio channel bitrate sample rate
```
Sami32: Audio channels, bitrate and sampling rate

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -1064,7 +1084,7 @@ public synchronized ProcessWrapper launchTranscode(
 					}
 				}
 
-				if (!customFFmpegOptions.contains("-ar ")) {
+				if (!customFFmpegOptions.contains("-ar ") && params.aid.getSampleFrequency() != params.mediaRenderer.getTranscodedVideoAudioSampleRate()) {
```
Sami32: Add params.aid != null to avoid NPE rising.

Sami32:
As it doesn't work i suggest something like that:
`".*-(crf|b|b:v|q|q:v|qscale|qscale:v) .*|((x264opts|x264-params) .*(crf|bitrate|b|q|qp)=).*"`

-x264params option1=23:option2=blabla:crf=18:option4=1

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -366,13 +367,15 @@
...
 				} else if (type() == FormatType.AUDIO) {
 					// Skip
-				} else if (renderer.isTranscodeToAAC()) {
-					transcodeOptions.add("-c:a");
-					transcodeOptions.add("aac");
 				} else {
-					if (!customFFmpegOptions.contains("-c:a ")) {
-						transcodeOptions.add("-c:a");
-						transcodeOptions.add("ac3");
+					if (!customFFmpegOptions.matches(".*-(c:a|codec:a|acodec) [^ ]+.*")) {
```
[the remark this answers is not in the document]
Nadahar: I can't evaluate a regex without knowing what it's intended to do. Is the goal here simply to catch `-c:a`, `-codec:a` and `-codeca`? If so, it can be simplified to something like:
```java
if (!Pattern.compile("-(?:c:a|codec:a|acodec)\\b").matcher(customFFmpegOptions).find() {
```

I rarely used `matches()` because it has the strange requirement that is has to match the entire string. That's often not what you want.

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -1055,7 +1075,7 @@ public synchronized ProcessWrapper launchTranscode(
 					cmdList.add(String.valueOf(channels));
 				}
 
-				if (!customFFmpegOptions.contains("-ab ")) {
+				if (!customFFmpegOptions.matches(".*-(ab|b:a|q:a).*")) {
```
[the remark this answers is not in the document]
Nadahar: I would avoid `matches()` like above.

src/main/java/net/pms/encoders/FFMpegVideo.java
Nadahar: `getTranscodedVideoAudioSampleRate()` (`TRANSCODED_VIDEO_AUDIO_SAMPLE_RATE `) is always ```48000``` except for Xbox 360. I really doubt it's needed to resample in so many cases, so I think we should consider making the default `-1` and then only set it for those that can only handle a certain sample rate.

Sami32: Sorry i should have said that it as bread time and still didn't fully review this line because i need more time.

src/main/java/net/pms/encoders/FFMpegVideo.java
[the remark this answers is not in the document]
Sami32: I suspected that you will said that about matcher use ;)

Yes, the only purpose is to catch such strings.

src/main/java/net/pms/encoders/FFMpegVideo.java
Sami32: This add is not wished, because i've removed it in this PR #93

Sami32: Well, with the actual state of the code i'll prefer stick to 44100 and 48000 by safety.

src/main/java/net/pms/encoders/FFMpegVideo.java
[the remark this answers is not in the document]
onon765trb: Thank for comments and ideas.

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -386,29 +389,43 @@
...
+
+					//Consider encoders which have "preset" option
+					if (!customFFmpegOptions.contains("-preset")) {
+						transcodeOptions.add("-preset");
+						transcodeOptions.add("ultrafast");
+					}
+
 				}
-				if (!customFFmpegOptions.contains("-preset")) {
+				if (!customFFmpegOptions.contains("-preset") && customFFmpegOptions.matches(".*-(libx264|libx265).*")) {
```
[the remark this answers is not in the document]
onon765trb: Thank for your pointing out. My typo.
All of h264 encoders do not have the option,`-preset`. It can be known by `ffmpeg -h encoder=`.
Errors happen if the option is automatically added.
My goal is to avoid it.

For MPEG-2, I do not worry because  the option,`-preset` is not added.

src/main/java/net/pms/encoders/FFMpegVideo.java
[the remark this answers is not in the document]
onon765trb: Thank you.

src/main/java/net/pms/encoders/FFMpegVideo.java
onon765trb: Thank you.

src/main/java/net/pms/encoders/FFMpegVideo.java
onon765trb: Thank you.

src/main/java/net/pms/encoders/FFMpegVideo.java
onon765trb: Thank you.

onon765trb:
Thank you.

src/main/java/net/pms/encoders/FFMpegVideo.java
Nadahar: @onon765trb I don't think we need to handle what we discuss above in this PR. That's existing issues that we can fix separately.

src/main/java/net/pms/encoders/FFMpegVideo.java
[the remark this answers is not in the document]
Sami32: Ok, i think i get your idea?
`libx26` for `-preset ultrafast` and
`h264_qsv|hevc_qsv|mpeg2_qsv` for `-preset veryfast`
I cannot remember the preset specificities for all these codecs variant.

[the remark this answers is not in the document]
Nadahar: @Sami32 A ghost?

src/main/java/net/pms/encoders/FFMpegVideo.java
onon765trb: I will delete it.

On the pitch problem
I experienced it when resampling audio.I never have done it since using the high quality resampler, soxr, not swr.
The implement of soxr option for DMS is completed.

onon765trb:
Sorry. 
Reexaming the options of libx264 and libx265, `-qscale` isn't supported. `-qp` is  supported.
Thank you for the opportunity of the reexamination.

Sami32:
`qscale` is a generic codec agnostic option so it is supported, but after looking deeper at it i see that a warning saying that "-crf" should be used instead. So `-qscale` it will not have any affect at all with libx26. codecs..
Looking deeper at the parameters now, i see that `q` shouldn't have been added as codecs internal parameter. I was sure that `b` and `bitrate` was accepted parameters but it seem that my remembering was wrong.
So what do you think about our revision? ;-)
`".*-(crf|b|b:v) .*|((x264opts|x264-params) .*(crf|qp)=).*"`
I didn't looked at how it should be writed to use with `Pattern.compile`.

Thanks for having make me look deeper at it :wink:

Nadahar:
@Sami32 
> I didn't looked at how it should be writed to use with `Pattern.compile()`

It's simpler/easier to write it without `matches()` because you only need to match the part you care about. I thus find it strange that you only did it the more complicated way :stuck_out_tongue_closed_eyes: 

I can't try to create an example though, because I don't understand what you're trying to match.

onon765trb:
Thank you for your trying. I should write my remaking.
It is
`".*-(crf|b|b:v|qp) .*|((x264opts|x264-params) .*(crf|bitrate|B|q|qp)=).*"`

Nadahar:
I don't quite understand the rules you're trying to make, but I guess that converting it to a "normal" regex would be something like this:
```regex
-(?:crf|b|b:v|qp)\b|(?:(?:x264opts|x264-params)\b.*(?:crf|bitrate|B|q|qp)=\S+)
```

It still doesn't look logical to me though, is it really the case that only the first variant will start with a `-`? I assume that there can be any number of other arguments betwwen `x264opts`/`264.params` and the `cfr`/`bitrate` etc part? In general there's no reason to use capturing groups when you don't need the results of the capture.

onon765trb:
> I assume that there can be any number of other arguments betwwen `x264opts`/`264.params` and the `cfr`/`bitrate` etc part? 

You are right. At first, I tried
`if (!customFFmpegOptions.matches(".*-(crf|b|b:v|q|q:v|(x264opts|x264-params) +([^ =]+=)*(crf|bitrate|B|q|qp)=).*"))`
This should be corrected as follows
`if (!customFFmpegOptions.matches(".*-(crf|b|b:v|(x264opts|x264-params) +([^ =]+=:)*(crf|bitrate|B|q|qp)=).*"))`

I though using `.*` is sure 

> In general there's no reason to use capturing groups when you don't need the results of the capture.
Does it mean that there a way to avoid duplicate options.

src/main/java/net/pms/encoders/FFMpegVideo.java
onon765trb: @Nadahar
I red PR #93. PR #93 does not avoid the needless option `-ar` when the sample rate of a input is the same as that DMS sets. This lines do so.

How do you think whether or not the needless option `-ar` should remain.

Nadahar: @onon765trb The removal of `-ar` when it's not needed is fine to keep, what I suggested we shouldn't do in this PR is all the other stuff me and @Sami32 started to discuss related to target sample rate. So, I think this is good as it is :+1: (given that any `null` checks that might be needed is added). If the earlier code already makes sure that it they can't be `null` (`params.aid` and `params.mediaRenderer`), there's no need to put a `null` check here though.

Nadahar:
First, I'd like to say that I think [regular-expressions.info](https://www.regular-expressions.info/reference.html) is an excellent reference that I still use if I'm unsure of the exact syntax.

I don't like Java's `matches()` method, because what it does is that it inserts a `^` before and a `$` after your regex. That means that the regex will have to match the complete string or line (depending on the parameters used). I dislike this for two reasons, one is that the regex becomes less readable/more difficult to understand because you have to match also the parts of the string you're not interested in. The second is performance. There's no need to ask regex to match a larger part of the string than you need.

It might be a bit confusing, but if you look at the code "behind the scenes" that is run when you use `matches()` it actually does a whole lot of stuff that's not needed.
This code is called:
```java
    public static boolean matches(String regex, CharSequence input) {
        Pattern p = Pattern.compile(regex);
        Matcher m = p.matcher(input);
        return m.matches();
    }
```
...which calls this code:
```java
    public boolean matches() {
        return match(from, ENDANCHOR);
    }
```
...which in turn calls this code:
```java
    boolean match(int from, int anchor) {
        this.hitEnd = false;
        this.requireEnd = false;
        from        = from < 0 ? 0 : from;
        this.first  = from;
        this.oldLast = oldLast < 0 ? from : oldLast;
        for (int i = 0; i < groups.length; i++)
            groups[i] = -1;
        acceptMode = anchor;
        boolean result = parentPattern.matchRoot.match(this, from, text);
        if (!result)
            this.first = -1;
        this.oldLast = this.last;
        return result;
    }
```

So, even though you don't see it, running `Pattern.compile(<regex>).matcher(<string>).find()` is more effective as it performs fewer operations, *and* it allows you to write a simpler and more logical regex that only has to match the part of the string you're interested in. When you use this, you can drop the `.*` in both ends of your regex, because they are there to match the rest of the string.

When `match()` or `find()` is called on a `Matcher` object, regex will also find and store the content of the "capture groups". The first capture group is always the entire match (which means the entire string if you use `matches()`). In addition, one capture group is captured for every group in your regex. A "group" in regex is anything between `(` and `)`. All these are then stored, even if you never use them. They can be retrieved with `matcher.group(<group index>)` if needed, but in the case of `matches()` you never even get access to the `Matcher` object.

Since we don't need the content of the groups in this case, you should use non-capturing groups. A non-capturing group in regex is written `(?:<group content>)`. That means that the content of this group won't be "captured" and stored separately while processing the regex. This is the reason I've added `?:` in my examples.

Sami32:
@onon735trb I'm sorry but i couldn't find any `q` and `bitrate` parameters for the libx26. codecs.
```
Encoder libx264 [libx264 H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10]:
    General capabilities: delay threads 
    Threading capabilities: auto
    Supported pixel formats: yuv420p yuvj420p yuv422p yuvj422p yuv444p yuvj444p nv12 nv16 nv21 yuv420p10le yuv422p10le yuv444p10le nv20le gray gray10le
libx264 AVOptions:
  -preset            <string>     E..V..... Set the encoding preset (cf. x264 --fullhelp) (default "medium")
  -tune              <string>     E..V..... Tune the encoding params (cf. x264 --fullhelp)
  -profile           <string>     E..V..... Set profile restrictions (cf. x264 --fullhelp) 
  -fastfirstpass     <boolean>    E..V..... Use fast settings when encoding first pass (default true)
  -level             <string>     E..V..... Specify level (as defined by Annex A)
  -passlogfile       <string>     E..V..... Filename for 2 pass stats
  -wpredp            <string>     E..V..... Weighted prediction for P-frames
  -a53cc             <boolean>    E..V..... Use A53 Closed Captions (if available) (default true)
  -x264opts          <string>     E..V..... x264 options
  -crf               <float>      E..V..... Select the quality for constant quality mode (from -1 to FLT_MAX) (default -1)
  -crf_max           <float>      E..V..... In CRF mode, prevents VBV from lowering quality beyond this point. (from -1 to FLT_MAX) (default -1)
  -qp                <int>        E..V..... Constant quantization parameter rate control method (from -1 to INT_MAX) (default -1)
  -aq-mode           <int>        E..V..... AQ method (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     variance                     E..V..... Variance AQ (complexity mask)
     autovariance                 E..V..... Auto-variance AQ
     autovariance-biased              E..V..... Auto-variance AQ with bias to dark scenes
  -aq-strength       <float>      E..V..... AQ strength. Reduces blocking and blurring in flat and textured areas. (from -1 to FLT_MAX) (default -1)
  -psy               <boolean>    E..V..... Use psychovisual optimizations. (default auto)
  -psy-trellis       <float>      E..V..... Specify psychovisual trellis (from -1 to FLT_MAX) (default -1)
  -psy-rd            <string>     E..V..... Strength of psychovisual optimization, in <psy-rd>:<psy-trellis> format.
  -rc-lookahead      <int>        E..V..... Number of frames to look ahead for frametype and ratecontrol (from -1 to INT_MAX) (default -1)
  -weightb           <boolean>    E..V..... Weighted prediction for B-frames. (default auto)
  -weightp           <int>        E..V..... Weighted prediction analysis method. (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     simple                       E..V.....
     smart                        E..V.....
  -ssim              <boolean>    E..V..... Calculate and print SSIM stats. (default auto)
  -intra-refresh     <boolean>    E..V..... Use Periodic Intra Refresh instead of IDR frames. (default auto)
  -bluray-compat     <boolean>    E..V..... Bluray compatibility workarounds. (default auto)
  -b-bias            <int>        E..V..... Influences how often B-frames are used (from INT_MIN to INT_MAX) (default INT_MIN)
  -b-pyramid         <int>        E..V..... Keep some B-frames as references. (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     strict                       E..V..... Strictly hierarchical pyramid
     normal                       E..V..... Non-strict (not Blu-ray compatible)
  -mixed-refs        <boolean>    E..V..... One reference per partition, as opposed to one reference per macroblock (default auto)
  -8x8dct            <boolean>    E..V..... High profile 8x8 transform. (default auto)
  -fast-pskip        <boolean>    E..V..... (default auto)
  -aud               <boolean>    E..V..... Use access unit delimiters. (default auto)
  -mbtree            <boolean>    E..V..... Use macroblock tree ratecontrol. (default auto)
  -deblock           <string>     E..V..... Loop filter parameters, in <alpha:beta> form.
  -cplxblur          <float>      E..V..... Reduce fluctuations in QP (before curve compression) (from -1 to FLT_MAX) (default -1)
  -partitions        <string>     E..V..... A comma-separated list of partitions to consider. Possible values: p8x8, p4x4, b8x8, i8x8, i4x4, none, all
  -direct-pred       <int>        E..V..... Direct MV prediction mode (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     spatial                      E..V.....
     temporal                     E..V.....
     auto                         E..V.....
  -slice-max-size    <int>        E..V..... Limit the size of each slice in bytes (from -1 to INT_MAX) (default -1)
  -stats             <string>     E..V..... Filename for 2 pass stats
  -nal-hrd           <int>        E..V..... Signal HRD information (requires vbv-bufsize; cbr not allowed in .mp4) (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     vbr                          E..V.....
     cbr                          E..V.....
  -avcintra-class    <int>        E..V..... AVC-Intra class 50/100/200 (from -1 to 200) (default -1)
  -me_method         <int>        E..V..... Set motion estimation method (from -1 to 4) (default -1)
     dia                          E..V.....
     hex                          E..V.....
     umh                          E..V.....
     esa                          E..V.....
     tesa                         E..V.....
  -motion-est        <int>        E..V..... Set motion estimation method (from -1 to 4) (default -1)
     dia                          E..V.....
     hex                          E..V.....
     umh                          E..V.....
     esa                          E..V.....
     tesa                         E..V.....
  -forced-idr        <boolean>    E..V..... If forcing keyframes, force them as IDR frames. (default false)
  -coder             <int>        E..V..... Coder type (from -1 to 1) (default default)
     default                      E..V.....
     cavlc                        E..V.....
     cabac                        E..V.....
     vlc                          E..V.....
     ac                           E..V.....
  -b_strategy        <int>        E..V..... Strategy to choose between I/P/B-frames (from -1 to 2) (default -1)
  -chromaoffset      <int>        E..V..... QP difference between chroma and luma (from INT_MIN to INT_MAX) (default -1)
  -sc_threshold      <int>        E..V..... Scene change threshold (from INT_MIN to INT_MAX) (default -1)
  -noise_reduction   <int>        E..V..... Noise reduction (from INT_MIN to INT_MAX) (default -1)
  -x264-params       <string>     E..V..... Override the x264 configuration using a :-separated list of key=value parameters
  -cqp               <int>        E..V..... Constant quantization parameter rate control method (from -1 to INT_MAX) (default -1)
  -aq_mode           <int>        E..V..... AQ method (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     variance                     E..V..... Variance AQ (complexity mask)
     autovariance                 E..V..... Auto-variance AQ
     autovariance-biased              E..V..... Auto-variance AQ with bias to dark scenes
  -aq_strength       <float>      E..V..... AQ strength. Reduces blocking and blurring in flat and textured areas. (from -1 to FLT_MAX) (default -1)
  -psy_rd            <string>     E..V..... Strength of psychovisual optimization, in <psy-rd>:<psy-trellis> format.
  -psy_trellis       <float>      E..V..... Specify psychovisual trellis (from -1 to FLT_MAX) (default -1)
  -rc_lookahead      <int>        E..V..... Number of frames to look ahead for frametype and ratecontrol (from -1 to INT_MAX) (default -1)
  -intra_refresh     <int>        E..V..... Use Periodic Intra Refresh instead of IDR frames. (from -1 to 1) (default -1)
  -bframebias        <int>        E..V..... Influences how often B-frames are used (from INT_MIN to INT_MAX) (default INT_MIN)
  -deblockalpha      <int>        E..V..... in-loop deblocking filter alphac0 parameter (from -6 to 6) (default 0)
  -deblockbeta       <int>        E..V..... in-loop deblocking filter beta parameter (from -6 to 6) (default 0)
  -complexityblur    <float>      E..V..... Reduce fluctuations in QP (before curve compression) (from -1 to FLT_MAX) (default -1)
  -directpred        <int>        E..V..... Direct MV prediction mode (from -1 to INT_MAX) (default -1)
     none                         E..V.....
     spatial                      E..V.....
     temporal                     E..V.....
     auto                         E..V.....
```
Could you tell where did you find them?

src/main/java/net/pms/encoders/FFMpegVideo.java
Sami32: Sound reasonnable to me. As long as the NPE is fixed it is fine for me if this code line is merged.

Sami32:
@onon765trb I hope that we didn't scared you with all our comments?
Feel free to tell us if you want we push commits into your PR.

onon765trb:
@Nadahar I will study your sample code. I was going to use only POSIX.

@Sami32 `x264opts|x264-params` is the options for usage of original `x264` options.
You can find them
```
x264 --fullhelp
x264 core:156 r2917 3cf8a9c
Syntax: x264 [options] -o outfile infile

Infile can be raw (in which case resolution is required),
  or YUV4MPEG (*.y4m),
  or Avisynth if compiled with support (yes).
  or libav* formats if compiled with lavf support (no) or ffms support (no).
Outfile type is selected by filename:
 .264 -> Raw bytestream
 .mkv -> Matroska
 .flv -> Flash Video
 .mp4 -> MP4 if compiled with GPAC or L-SMASH support (no)
Output bit depth: 8/10
.
Options:

  -h, --help                  List basic options
      --longhelp              List more options
      --fullhelp              List all options

Example usage:

      Constant quality mode:
            x264 --crf 24 -o <output> <input>

      Two-pass with a bitrate of 1000kbps:
            x264 --pass 1 --bitrate 1000 -o <output> <input>
            x264 --pass 2 --bitrate 1000 -o <output> <input>

      Lossless:
            x264 --qp 0 -o <output> <input>

      Maximum PSNR at the cost of speed and visual quality:
            x264 --preset placebo --tune psnr -o <output> <input>

      Constant bitrate at 1000kbps with a 2 second-buffer:
            x264 --vbv-bufsize 2000 --bitrate 1000 -o <output> <input>

Presets:

      --profile <string>      Force the limits of an H.264 profile
                                  Overrides all settings.
                                  - baseline:
                                    --no-8x8dct --bframes 0 --no-cabac
                                    --cqm flat --weightp 0
                                    No interlaced.
                                    No lossless.
                                  - main:
                                    --no-8x8dct --cqm flat
                                    No lossless.
                                  - high:
                                    No lossless.
                                  - high10:
                                    No lossless.
                                    Support for bit depth 8-10.
                                  - high422:
                                    No lossless.
                                    Support for bit depth 8-10.
                                    Support for 4:2:0/4:2:2 chroma subsampling.
                                  - high444:
                                    Support for bit depth 8-10.
                                    Support for 4:2:0/4:2:2/4:4:4 chroma subsampling.
      --preset <string>       Use a preset to select encoding settings [medium]
                                  Overridden by user settings.
                                  - ultrafast:
                                    --no-8x8dct --aq-mode 0 --b-adapt 0
                                    --bframes 0 --no-cabac --no-deblock
                                    --no-mbtree --me dia --no-mixed-refs
                                    --partitions none --rc-lookahead 0 --ref 1
                                    --scenecut 0 --subme 0 --trellis 0
                                    --no-weightb --weightp 0
                                  - superfast:
                                    --no-mbtree --me dia --no-mixed-refs
                                    --partitions i8x8,i4x4 --rc-lookahead 0
                                    --ref 1 --subme 1 --trellis 0 --weightp 1
                                  - veryfast:
                                    --no-mixed-refs --rc-lookahead 10
                                    --ref 1 --subme 2 --trellis 0 --weightp 1
                                  - faster:
                                    --no-mixed-refs --rc-lookahead 20
                                    --ref 2 --subme 4 --weightp 1
                                  - fast:
                                    --rc-lookahead 30 --ref 2 --subme 6
                                    --weightp 1
                                  - medium:
                                    Default settings apply.
                                  - slow:
                                    --direct auto --rc-lookahead 50 --ref 5
                                    --subme 8 --trellis 2
                                  - slower:
                                    --b-adapt 2 --direct auto --me umh
                                    --partitions all --rc-lookahead 60
                                    --ref 8 --subme 9 --trellis 2
                                  - veryslow:
                                    --b-adapt 2 --bframes 8 --direct auto
                                    --me umh --merange 24 --partitions all
                                    --ref 16 --subme 10 --trellis 2
                                    --rc-lookahead 60
                                  - placebo:
                                    --bframes 16 --b-adapt 2 --direct auto
                                    --slow-firstpass --no-fast-pskip
                                    --me tesa --merange 24 --partitions all
                                    --rc-lookahead 60 --ref 16 --subme 11
                                    --trellis 2
      --tune <string>         Tune the settings for a particular type of source
                              or situation
                                  Overridden by user settings.
                                  Multiple tunings are separated by commas.
                                  Only one psy tuning can be used at a time.
                                  - film (psy tuning):
                                    --deblock -1:-1 --psy-rd <unset>:0.15
                                  - animation (psy tuning):
                                    --bframes {+2} --deblock 1:1
                                    --psy-rd 0.4:<unset> --aq-strength 0.6
                                    --ref {Double if >1 else 1}
                                  - grain (psy tuning):
                                    --aq-strength 0.5 --no-dct-decimate
                                    --deadzone-inter 6 --deadzone-intra 6
                                    --deblock -2:-2 --ipratio 1.1
                                    --pbratio 1.1 --psy-rd <unset>:0.25
                                    --qcomp 0.8
                                  - stillimage (psy tuning):
                                    --aq-strength 1.2 --deblock -3:-3
                                    --psy-rd 2.0:0.7
                                  - psnr (psy tuning):
                                    --aq-mode 0 --no-psy
                                  - ssim (psy tuning):
                                    --aq-mode 2 --no-psy
                                  - fastdecode:
                                    --no-cabac --no-deblock --no-weightb
                                    --weightp 0
                                  - zerolatency:
                                    --bframes 0 --force-cfr --no-mbtree
                                    --sync-lookahead 0 --sliced-threads
                                    --rc-lookahead 0
      --slow-firstpass        Don't force these faster settings with --pass 1:
                                  --no-8x8dct --me dia --partitions none
                                  --ref 1 --subme {2 if >2 else unchanged}
                                  --trellis 0 --fast-pskip

Frame-type options:

  -I, --keyint <integer or "infinite"> Maximum GOP size [250]
  -i, --min-keyint <integer>  Minimum GOP size [auto]
      --no-scenecut           Disable adaptive I-frame decision
      --scenecut <integer>    How aggressively to insert extra I-frames [40]
      --intra-refresh         Use Periodic Intra Refresh instead of IDR frames
  -b, --bframes <integer>     Number of B-frames between I and P [3]
      --b-adapt <integer>     Adaptive B-frame decision method [1]
                                  Higher values may lower threading efficiency.
                                  - 0: Disabled
                                  - 1: Fast
                                  - 2: Optimal (slow with high --bframes)
      --b-bias <integer>      Influences how often B-frames are used [0]
      --b-pyramid <string>    Keep some B-frames as references [normal]
                                  - none: Disabled
                                  - strict: Strictly hierarchical pyramid
                                  - normal: Non-strict (not Blu-ray compatible)
      --open-gop              Use recovery points to close GOPs
                              Only available with b-frames
      --no-cabac              Disable CABAC
  -r, --ref <integer>         Number of reference frames [3]
      --no-deblock            Disable loop filter
  -f, --deblock <alpha:beta>  Loop filter parameters [0:0]
      --slices <integer>      Number of slices per frame; forces rectangular
                              slices and is overridden by other slicing options
      --slices-max <integer>  Absolute maximum slices per frame; overrides
                              slice-max-size/slice-max-mbs when necessary
      --slice-max-size <integer> Limit the size of each slice in bytes
      --slice-max-mbs <integer> Limit the size of each slice in macroblocks (max)
      --slice-min-mbs <integer> Limit the size of each slice in macroblocks (min)
      --tff                   Enable interlaced mode (top field first)
      --bff                   Enable interlaced mode (bottom field first)
      --constrained-intra     Enable constrained intra prediction.
      --pulldown <string>     Use soft pulldown to change frame rate
                                  - none, 22, 32, 64, double, triple, euro (requires cfr input)
      --fake-interlaced       Flag stream as interlaced but encode progressive.
                              Makes it possible to encode 25p and 30p Blu-Ray
                              streams. Ignored in interlaced mode.
      --frame-packing <integer> For stereoscopic videos define frame arrangement
                                  - 0: checkerboard - pixels are alternatively from L and R
                                  - 1: column alternation - L and R are interlaced by column
                                  - 2: row alternation - L and R are interlaced by row
                                  - 3: side by side - L is on the left, R on the right
                                  - 4: top bottom - L is on top, R on bottom
                                  - 5: frame alternation - one view per frame
                                  - 6: mono - 2D frame without any frame packing
                                  - 7: tile format - L is on top-left, R split across

Ratecontrol:

  -q, --qp <integer>          Force constant QP (0-81, 0=lossless)
  -B, --bitrate <integer>     Set bitrate (kbit/s)
      --crf <float>           Quality-based VBR (-12-51) [23.0]
      --rc-lookahead <integer> Number of frames for frametype lookahead [40]
      --vbv-maxrate <integer> Max local bitrate (kbit/s) [0]
      --vbv-bufsize <integer> Set size of the VBV buffer (kbit) [0]
      --vbv-init <float>      Initial VBV buffer occupancy [0.9]
      --crf-max <float>       With CRF+VBV, limit RF to this value
                                  May cause VBV underflows!
      --qpmin <integer>       Set min QP [0]
      --qpmax <integer>       Set max QP [81]
      --qpstep <integer>      Set max QP step [4]
      --ratetol <float>       Tolerance of ABR ratecontrol and VBV [1.0]
      --ipratio <float>       QP factor between I and P [1.40]
      --pbratio <float>       QP factor between P and B [1.30]
      --chroma-qp-offset <integer>  QP difference between chroma and luma [0]
      --aq-mode <integer>     AQ method [1]
                                  - 0: Disabled
                                  - 1: Variance AQ (complexity mask)
                                  - 2: Auto-variance AQ
                                  - 3: Auto-variance AQ with bias to dark scenes
      --aq-strength <float>   Reduces blocking and blurring in flat and
                              textured areas. [1.0]

  -p, --pass <integer>        Enable multipass ratecontrol
                                  - 1: First pass, creates stats file
                                  - 2: Last pass, does not overwrite stats file
                                  - 3: Nth pass, overwrites stats file
      --stats <string>        Filename for 2 pass stats ["x264_2pass.log"]
      --no-mbtree             Disable mb-tree ratecontrol.
      --qcomp <float>         QP curve compression [0.60]
      --cplxblur <float>      Reduce fluctuations in QP (before curve compression) [20.0]
      --qblur <float>         Reduce fluctuations in QP (after curve compression) [0.5]
      --zones <zone0>/<zone1>/...  Tweak the bitrate of regions of the video
                              Each zone is of the form
                                  <start frame>,<end frame>,<option>
                                  where <option> is either
                                      q=<integer> (force QP)
                                  or  b=<float> (bitrate multiplier)
      --qpfile <string>       Force frametypes and QPs for some or all frames
                              Format of each line: framenumber frametype QP
                              QP is optional (none lets x264 choose). Frametypes: I,i,K,P,B,b.
                                  K=<I or i> depending on open-gop setting
                              QPs are restricted by qpmin/qpmax.

Analysis:

  -A, --partitions <string>   Partitions to consider ["p8x8,b8x8,i8x8,i4x4"]
                                  - p8x8, p4x4, b8x8, i8x8, i4x4
                                  - none, all
                                  (p4x4 requires p8x8. i8x8 requires --8x8dct.)
      --direct <string>       Direct MV prediction mode ["spatial"]
                                  - none, spatial, temporal, auto
      --no-weightb            Disable weighted prediction for B-frames
      --weightp <integer>     Weighted prediction for P-frames [2]
                                  - 0: Disabled
                                  - 1: Weighted refs
                                  - 2: Weighted refs + Duplicates
      --me <string>           Integer pixel motion estimation method ["hex"]
                                  - dia: diamond search, radius 1 (fast)
                                  - hex: hexagonal search, radius 2
                                  - umh: uneven multi-hexagon search
                                  - esa: exhaustive search
                                  - tesa: hadamard exhaustive search (slow)
      --merange <integer>     Maximum motion vector search range [16]
      --mvrange <integer>     Maximum motion vector length [-1 (auto)]
      --mvrange-thread <int>  Minimum buffer between threads [-1 (auto)]
  -m, --subme <integer>       Subpixel motion estimation and mode decision [7]
                                  - 0: fullpel only (not recommended)
                                  - 1: SAD mode decision, one qpel iteration
                                  - 2: SATD mode decision
                                  - 3-5: Progressively more qpel
                                  - 6: RD mode decision for I/P-frames
                                  - 7: RD mode decision for all frames
                                  - 8: RD refinement for I/P-frames
                                  - 9: RD refinement for all frames
                                  - 10: QP-RD - requires trellis=2, aq-mode>0
                                  - 11: Full RD: disable all early terminations
      --psy-rd <float:float>  Strength of psychovisual optimization ["1.0:0.0"]
                                  #1: RD (requires subme>=6)
                                  #2: Trellis (requires trellis, experimental)
      --no-psy                Disable all visual optimizations that worsen
                              both PSNR and SSIM.
      --no-mixed-refs         Don't decide references on a per partition basis
      --no-chroma-me          Ignore chroma in motion estimation
      --no-8x8dct             Disable adaptive spatial transform size
  -t, --trellis <integer>     Trellis RD quantization. [1]
                                  - 0: disabled
                                  - 1: enabled only on the final encode of a MB
                                  - 2: enabled on all mode decisions
      --no-fast-pskip         Disables early SKIP detection on P-frames
      --no-dct-decimate       Disables coefficient thresholding on P-frames
      --nr <integer>          Noise reduction [0]

      --deadzone-inter <int>  Set the size of the inter luma quantization deadzone [21]
      --deadzone-intra <int>  Set the size of the intra luma quantization deadzone [11]
                                  Deadzones should be in the range 0 - 32.
      --cqm <string>          Preset quant matrices ["flat"]
                                  - jvt, flat
      --cqmfile <string>      Read custom quant matrices from a JM-compatible file
                                  Overrides any other --cqm* options.
      --cqm4 <list>           Set all 4x4 quant matrices
                                  Takes a comma-separated list of 16 integers.
      --cqm8 <list>           Set all 8x8 quant matrices
                                  Takes a comma-separated list of 64 integers.
      --cqm4i, --cqm4p, --cqm8i, --cqm8p <list>
                              Set both luma and chroma quant matrices
      --cqm4iy, --cqm4ic, --cqm4py, --cqm4pc <list>
                              Set individual quant matrices

Video Usability Info (Annex E):
The VUI settings are not used by the encoder but are merely suggestions to
the playback equipment. See doc/vui.txt for details. Use at your own risk.

      --overscan <string>     Specify crop overscan setting ["undef"]
                                  - undef, show, crop
      --videoformat <string>  Specify video format ["undef"]
                                  - component, pal, ntsc, secam, mac, undef
      --range <string>        Specify color range ["auto"]
                                  - auto, tv, pc
      --colorprim <string>    Specify color primaries ["undef"]
                                  - undef, bt709, bt470m, bt470bg, smpte170m,
                                    smpte240m, film, bt2020, smpte428,
                                    smpte431, smpte432
      --transfer <string>     Specify transfer characteristics ["undef"]
                                  - undef, bt709, bt470m, bt470bg, smpte170m,
                                    smpte240m, linear, log100, log316,
                                    iec61966-2-4, bt1361e, iec61966-2-1,
                                    bt2020-10, bt2020-12, smpte2084, smpte428,
                                    arib-std-b67
      --colormatrix <string>  Specify color matrix setting ["???"]
                                  - undef, bt709, fcc, bt470bg, smpte170m,
                                    smpte240m, GBR, YCgCo, bt2020nc, bt2020c,
                                    smpte2085, chroma-derived-nc,
                                    chroma-derived-c, ICtCp
      --chromaloc <integer>   Specify chroma sample location (0 to 5) [0]
      --alternative-transfer <string> Specify an alternative transfer
                              characteristics ["undef"]
                                  - same values as --transfer
      --nal-hrd <string>      Signal HRD information (requires vbv-bufsize)
                                  - none, vbr, cbr (cbr not allowed in .mp4)
      --filler                Force hard-CBR and generate filler (implied by
                              --nal-hrd cbr)
      --pic-struct            Force pic_struct in Picture Timing SEI
      --crop-rect <string>    Add 'left,top,right,bottom' to the bitstream-level
                              cropping rectangle

Input/Output:

  -o, --output <string>       Specify output file
      --muxer <string>        Specify output container format ["auto"]
                                  - auto, raw, mkv, flv
      --demuxer <string>      Specify input container format ["auto"]
                                  - auto, raw, y4m, avs
      --input-fmt <string>    Specify input file format (requires lavf support)
      --input-csp <string>    Specify input colorspace format for raw input
                              - valid csps for `raw' demuxer:
                                i400, i420, yv12, nv12, nv21, i422, yv16, nv16, 
                                yuyv, uyvy, i444, yv24, bgr, bgra, rgb
      --output-csp <string>   Specify output colorspace ["i400"]
                                  - i400, i420, i422, i444, rgb
      --input-depth <integer> Specify input bit depth for raw input
      --output-depth <integer> Specify output bit depth
      --input-range <string>  Specify input color range ["auto"]
                                  - auto, tv, pc
      --input-res <intxint>   Specify input resolution (width x height)
      --index <string>        Filename for input index file
      --sar width:height      Specify Sample Aspect Ratio
      --fps <float|rational>  Specify framerate
      --seek <integer>        First frame to encode
      --frames <integer>      Maximum number of frames to encode
      --level <string>        Specify level (as defined by Annex A)
      --bluray-compat         Enable compatibility hacks for Blu-ray support
      --avcintra-class <integer> Use compatibility hacks for AVC-Intra class
                                  - 50, 100, 200
      --stitchable            Don't optimize headers based on video content
                              Ensures ability to recombine a segmented encode

  -v, --verbose               Print stats for each frame
      --no-progress           Don't show the progress indicator while encoding
      --quiet                 Quiet Mode
      --log-level <string>    Specify the maximum level of logging ["info"]
                                  - none, error, warning, info, debug
      --psnr                  Enable PSNR computation
      --ssim                  Enable SSIM computation
      --threads <integer>     Force a specific number of threads
      --lookahead-threads <integer> Force a specific number of lookahead threads
      --sliced-threads        Low-latency but lower-efficiency threading
      --thread-input          Run Avisynth in its own thread
      --sync-lookahead <integer> Number of buffer frames for threaded lookahead
      --non-deterministic     Slightly improve quality of SMP, at the cost of repeatability
      --cpu-independent       Ensure exact reproducibility across different cpus,
                                  as opposed to letting them select different algorithms
      --asm <integer>         Override CPU detection
      --no-asm                Disable all CPU optimizations
      --opencl                Enable use of OpenCL
      --opencl-clbin <string> Specify path of compiled OpenCL kernel cache
      --opencl-device <integer> Specify OpenCL device ordinal
      --dump-yuv <string>     Save reconstructed frames
      --sps-id <integer>      Set SPS and PPS id numbers [0]
      --aud                   Use access unit delimiters
      --force-cfr             Force constant framerate timestamp generation
      --tcfile-in <string>    Force timestamp generation with timecode file
      --tcfile-out <string>   Output timecode v2 file from input timestamps
      --timebase <int/int>    Specify timebase numerator and denominator
                 <integer>    Specify timebase numerator for input timecode file
                              or specify timebase denominator for other input
      --dts-compress          Eliminate initial delay with container DTS hack

Filtering:

      --vf, --video-filter <filter0>/<filter1>/... Apply video filtering to the input file

      Filter options may be specified in <filter>:<option>=<value> format.

      Available filters:
      crop:left,top,right,bottom
            removes pixels from the edges of the frame
      resize:[width,height][,sar][,fittobox][,csp][,method]
            resizes frames based on the given criteria:
            - resolution only: resizes and adapts sar to avoid stretching
            - sar only: sets the sar and resizes to avoid stretching
            - resolution and sar: resizes to given resolution and sets the sar
            - fittobox: resizes the video based on the desired constraints
               - width, height, both
            - fittobox and sar: same as above except with specified sar
            - csp: convert to the given csp. syntax: [name][:depth]
               - valid csp names [keep current]: i400, i420, yv12, nv12, nv21, i422, yv16, nv16, yuyv, uyvy, i444, yv24, bgr, bgra, rgb
               - depth: 8 or 16 bits per pixel [keep current]
            note: not all depths are supported by all csps.
            - method: use resizer method ["bicubic"]
               - fastbilinear, bilinear, bicubic, experimental, point,
               - area, bicublin, gauss, sinc, lanczos, spline
      select_every:step,offset1[,...]
            apply a selection pattern to input frames
            step: the number of frames in the pattern
            offsets: the offset into the step to select a frame
            see: http://avisynth.nl/index.php/Select#SelectEvery
```
And I am preparing for the improvements of `x264` options.

Sami32:
@onon735trb x264 excecutable is not FFmpeg, we doesn't use it. And yes their options and parameters are different.

Did i missed something?

onon765trb:
You can find them in the below.
`` ` 
 Ratecontrol:

  -q, --qp <integer>          Force constant QP (0-81, 0=lossless)
  -B, --bitrate <integer>     Set bitrate (kbit/s)
`` `

Sami32:
@onon765trb It seem we don't understand each other. These parameter doesn't come from FFmpeg and we only use FFmpeg.

You seem refered to the VLC executable.

onon765trb:
Rather than it, it is a lack of my confirmation and tests. `-x264opts` and `-x264-params` are options to use x264 options, as show by  [16.8.2 Options](https://ffmpeg.org/ffmpeg-all.html#Options-26). But, short cut options `-q` and `-B` are not parsed by ffmpeg.
You can confirm by
```
ffmpeg -i input -c:v libx264 -x264-params q=21 -f null -
ffmpeg -i input -c:v libx264 -x264-params B=3000 -f null -
ffmpeg -i input -c:v libx264 -x264-params bitrate=3000 -f null -
ffmpeg -i input -c:v libx264 -x264opts q=21 -f null -
ffmpeg -i input -c:v libx264 -x264opts B=3000 -f null -
ffmpeg -i input -c:v libx264 -x264opts bitrate=3000 -f null -
```

By the way, there wa what I noticed when I reviewed my revision.
```
					if (!customFFmpegOptions.contains("-tune")) {
						transcodeOptions.add("-tune");
						transcodeOptions.add("zerolatency");
					}

					//Consider encoders which have "preset" option
					if (!customFFmpegOptions.contains("-preset")) {
						transcodeOptions.add("-preset");
						transcodeOptions.add("ultrafast");
					}

				}
				if (!customFFmpegOptions.contains("-preset") && customFFmpegOptions.matches(".*-(libx264|libx265).*")) {
					transcodeOptions.add("-preset");
					transcodeOptions.add("ultrafast");
				}
				if (!customFFmpegOptions.contains("-level")) {
					transcodeOptions.add("-level");
					transcodeOptions.add("31");
				}
```
Should this be placed in `public List<String> getVideoBitrateOptions` ? Very strange.
I look FFmpegWebVideo.java and WebRender.java. It seems to me that this change don't have effects on them.

In `if (!(renderer instanceof RendererConfiguration.OutputOverride) && configuration.isFFmpegMuxWithTsMuxerWhenCompatible())` on line 994, the existance of tsmuxer is not check by `PlayerFactory.isPlayerActive(TsMuxeRVideo.ID)`. Isn't this a cause for errors?

Sami32:
@Ono765trb Yes, because they are not FFmpeg parameters. But i agree with you that their documentation refering to the VLC x264 to get more informations is weird and unclear. They should have removed that or added that it concerned only the documentation related to the same FFmpeg parameters only.
At least now i understand why you have set them and where you get them ;)

Yes i know. It was only MEncoder hardcoded settings only used as a hack for XBOX in 2009, then UMS in around 2013 make `getVideoBitrateOptions` useable from renderer's configuration and added it to FFmpegVideo only.
It is why i want totally remove that XBOX hack and clean the code.

Don't worry for VLCVideo.java, it can be done in an other PR if need.
VLC use x264 but have its own way for the settings AFAIK.
https://www.videolan.org/developers/x264.html
I never really looked at that but i know that their settings are writed differently than x264.
https://wiki.videolan.org/VLC_command-line_help/

Yes, you are right that defering to tsMuxeR without checking PlayerFactory.isPlayerActive(TsMuxeRVideo.ID) is a source of issues waiting to happen.
And my poor Java skill doesn't see this check done anywhere.

Sami32:
@onon765trb Yes you are right that `bitrate` work, i also didn't tested :+1:

src/main/java/net/pms/encoders/FFMpegVideo.java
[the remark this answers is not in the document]
Sami32: We should probably add `-vpre` as well.

Sami32:
As a compromise i suggest to add in our last wizard window related to the library folders choice to add a line asking users if they want add the "Desktop" (i'll not do any objection if this one was removed) and the "Download" folder by default.
As users will mostly hit "Yes" without reading the message as we already saw many times, it should be fine to avoid complain that they don't find they media and from the few other wishing not having them added for burding issue as these last ones should take the time to read the message.

Nadahar:
@Sami32 I assume you meant to post your comment in #101?

Sami32:
@nadahar Oups...yes 8/

onon765trb:
On line 546
```
			final boolean isTsMuxeRVideoEngineActive = PlayerFactory.isPlayerActive(TsMuxeRVideo.ID);

			// Output audio codec
			dtsRemux = isTsMuxeRVideoEngineActive &&
```

On line 996
```
			if (deferToTsmuxer && !PlayerFactory.isPlayerAvailable(StandardPlayerId.TSMUXER_VIDEO)) {
```

The inconsistency have reasons? tsmuxer player is enabled or disabled.

Now I am fixing potential error found newly.

Nadahar:
@onon765trb 
> The inconsistency have reasons? 

I don't think there are any other reason for this than that the code isn't very organized and that these tests have been added at different times. Feel free to fix it :wink:

Sami32:
@onon765trb Perhaps i missed something but i don't see inconsistency here.
It first check if the DTSremux parameters should be applied that will be handled later by tsMuxeR engine.
Then before the transcoding it make sure that the media is video and not audio type.

onon765trb:
My understanding is 
`PlayerFactory.isPlayerActive` is that a TsMuxeR binary exists and TsMuxeR Player is enabled.
`PlayerFactory.isPlayerAvailable` is that only a TsMuxeR binary exists.

Misunderstanding? I do not mind which is better.
If both of you have requests, tell me. I will fix.

Sami32:
@onon765trb You seem right, `isPlayerActive` should be prefered.

Nadahar:
@onon765trb You are correct, except that `PlayerFactory.isPlayerAvailable` checks a bit more than that the binary exists. It also means that some basic command like `--version` (or similar) have been run and gave the expected result, so that the binary actually "works" in the sense that it's a valid executable for the platform/bitness and that there's no missing dependencies or other circumstances that prevents it from running.

`PlayerFactory.isPlayerAvailable` simply means that the engine/player is available for DMS to use, `PlayerFactory.isPlayerEnabled` means that the engine/player is enabled by the user in the configuration and `PlayerFactory.isPlayerActive` means that it is both available and enabled.

Sami32:
@Onon765trb Yes, the Web render code has been weirdly added as a separated part years ago, that is why you don't get these informations.
But as it will probably a full refactoring of the Web part i think that it should deserve a separated PR later.
No ETA.

PR #107, opened 2018-10-17
Modify wizard to ask to share default folders

### LINKED ISSUE #101: Auto sharing folder on wizard
DMS automatically share Desktop and Downloads folders.
I think this is not good on security and the burdens of PC.
Desktop and Downloads folders are privacy itself.
It is a cause for unexpected outflows by beginners.

And I think agreements by users are need, including what folders is shared, when automatically sharing folders,

Nadahar:
This is just a suggestion to resolve #101, the code hasn't been completed or cleaned up. I've disabled the other "wizard" questions for testing speed, and the dialog doesn't actually do anything yet. GUI design is so slow that I want to come to an agreement of the concept before I do the rest of the implementation.

Remains to be done:
- [ ] Clean up not used code, uncomment disabled code
- [ ] Implement translatable texts
- [ ] Fix the title and number of "questions"
- [ ] Actually set the setting according to the choice made

@Sami32 @onon765trb What do you think about the solution?

### DIFF
--- a/src/main/java/net/pms/newgui/ConfigurationWizard.java
+++ b/src/main/java/net/pms/newgui/ConfigurationWizard.java
@@ -18,15 +18,36 @@
  */
 package net.pms.newgui;
 
+import java.awt.Dimension;
+import java.awt.Font;
+import java.awt.FontMetrics;
+import java.awt.GridBagConstraints;
+import java.awt.GridBagLayout;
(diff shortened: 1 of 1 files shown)


Sami32:
Look perfect to me.

onon765trb:
I will try tests.

PR #112, opened 2018-10-25
M4B, M4P and M4R

Sami32:
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/112)


src/main/java/net/pms/formats/audio/M4A.java
```diff
@@ -27,6 +27,11 @@ public Identifier getIdentifier() {
 	 */
 	@Override
 	public String[] getSupportedExtensions() {
-		return new String[] { "m4a" };
+		return new String[] {
+			"m4a",
+			"m4b",
+			"m4p",
```
Nadahar:
```suggestion
```

Nadahar: I don't see any point in including these, as they are encrypted and can only be decrypted by iTunes. If they are decrypted they are simply M4A.

src/test/java/net/pms/test/formats/FormatFactoryTest.java
```diff
@@ -114,6 +114,9 @@ public final void testFormatRetrieval() {
 		testSingleFormat("test.jpg", "JPG", FormatType.IMAGE);
 		testSingleFormat("test.mpo", "JPG", FormatType.IMAGE);
 		testSingleFormat("test.m4a", "M4A", FormatType.AUDIO);
+		testSingleFormat("test.m4b", "M4A", FormatType.AUDIO);
+		testSingleFormat("test.m4p", "M4A", FormatType.AUDIO);
```
Nadahar:
```suggestion
```

Nadahar: The same applies here.

src/main/java/net/pms/formats/audio/M4A.java
Sami32: The point is only for macOS users to be able to stream them, also if using FFmpeg parsing

Nadahar: But they are encrypted.... they can't stream them.

Sami32: The encryption could have been removed and the extension not changed.
But now users can play them also if their audio media have Apple FairPlay DRM because they can configure until five renderers to play them.

Sami32:
https://developer.apple.com/streaming/fps/
https://en.wikipedia.org/wiki/FairPlay

I guess that @Onon765trb could probably confirm that?

Sami32:
Yes, only from renderers having iTunes installed.

src/main/java/net/pms/formats/audio/M4A.java
Nadahar: I doubt DMS can be "configured as a renderer" without supporting their DRM scheme. When it comes to removing the encryption and not renaming them, I find that highly unlikely. I'm pretty sure that if it's possible to remove the encryption with some Apple software, it will also change the extension.

I only see this as very likely to just create problems with files that won't play.

Sami32:
Removing FairPlay is not done using Apple software Lol

Nadahar:
> Removing FairPlay is not done using Apple software

As I suspected, as they abandoned FairPlay it's reasonable that they can remove it from already purchased files. I'm pretty sure that the M4P extension is removed then as well, and this this is pretty much a non-issue today.

Sami32:
All the Apple TV and iPhone, iPad should be able to play them .

Sami32:
https://support.apple.com/fr-ci/HT202190

Doesn't DMS able to use the iTunes libraries?

Nadahar:
Using the iTunes libraries and implementing Apple's DRM is two very different things.

Sami32:
That is said i could be wrong about Itunes libraries handling by DMS.

Nadahar:
I doubt that we can even parse them, but I haven't found a sample file so I can't be sure. It depends on if they have encrypted the header as well. If the header is encrypted, we have no way to handle them, and if they aren't this will still produce errors on everything but Apple renderers. 

I think this is an issue to be solved by those that have chosen to buy music with DRM, not us.

Sami32:
___
*[src/main/java/net/pms/formats/audio/M4A.java, line 33 at r1](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/112#-LPgyT4tEVQQt6rhJTvs-r1-33:-LQZ-hne1VEjbKr5QIBo:b-896fix) ([raw file](https://github.com/digitalmediaserver/digitalmediaserver/blob/52b83a663d194f59a268cba551353ac94177d182/src/main/java/net/pms/formats/audio/M4A.java#L33)):*

I doubt DMS can be "configured as a renderer" without supporting their DRM scheme. When it comes to removing the encryption and not renaming them, I find that highly unlikely. I'm pretty sure that if it's possible to remove the encryption with some Apple software, it will also change the extension.

I only see this as very likely to just create problems with files that won't play.
</blockquote></details>

Done.

Sami32:
Hehe, thanks :wink:

Sorry, i couldn't resist to add some missing extensions after my own review ;)

Nadahar:
I know :stuck_out_tongue_closed_eyes:

PR #115, opened 2018-10-26
Video level limits

Nadahar:
I'm trying to salvage our previous work on H.264 and H.265 level. However, there are some stuff here that really doesn't make sense to me, so I need to ask some questions before I continue.

---
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/115)


src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1676,115 +1683,104 @@ public void postParse(InputFile f) {
...
+			videoWithinH264Level41Limits = true;
+			if (
+				container != null &&
+				(
+					container.equals("matroska") ||
+					container.equals("mkv") ||
+					container.equals("mov") ||
+					container.equals("mp4")
+				)
+			) { // Containers without h264_annexB
```
Nadahar: @Sami32 What is all this annex B stuff good for? Parsing it like this (by opening the file and interpreting the bytes) seems to me like a repeat of the parsing that has already been done. Does it serve a purpose? Why isn't the parsed "level" good enough?

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1676,115 +1683,104 @@ public void postParse(InputFile f) {
...
+										 * Meaning this math is correct
+										 */
+										maxref = (int) Math.floor(10252743 / (double) (width * height));
+									} else {
+										/*
+										 * This is the math for level 4.1, which results in:
+										 *    - 4 for 1920x1080
+										 *    - 9 for 1280x720
+										 */
+										maxref = (int) Math.floor(8388608 / (double) (width * height));
```
Nadahar: The above code looks like some kind of manual level calculation to try to make a specific, non-standardized level for the PS3. Why can't we just use level 4.2 instead?

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Nadahar: This code doesn't remove anything from the file..?

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: That is said it was years ago and perhaps PS3 get new firmware updates that fixed it.

Nadahar:
Now I'm starting to understand. This has been completely fucked up by 4b4699f7aa9ba38f7abf4f424b1de97eb75c6448 :rofl:

Basically, the attempt to cherry pick the referenced commits have failed and the code has been rendered pretty much useless:
https://github.com/ps3mediaserver/ps3mediaserver/commit/8490dbcc7a586b570fd3d15f6dc38a9fe0e18776
https://github.com/ps3mediaserver/ps3mediaserver/commit/790a7bc710f5b974615279387aec6b0116a08119
https://github.com/ps3mediaserver/ps3mediaserver/commit/d6993d238b7b07dab94548053b5d7435a45addc0

The old code actually used FFmpeg to generate a frame with annex B, and then the header from this frame was parsed and the reference count determined. After the flawed commit, the frame is still extracted with FFmpeg but isn't used for anything :astonished:

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: I will ssid more, this code introduce some piping issue while used by tsMuxeR.
I'm in favor of removing all this annex B code.

Sami32: And to be honest that code is only here because the guy that encoded the media didn't leaved the default and set non standard value. I guess that it was most probable 10 years ago.

Sami32: FFmpeg and MediaInfo are able to give that information.
I think that was because in that time many video was unstandard transcoded and some software or users set the level to a faked 5.1, so that make many renderers transcode them on a 10 years old computer...

Sami32:
Kept this code if you want but the math are flawed, as they only consider
the video level of 4.1 and an HD resolution. So as soon as you run outside
of these limited cases the math get wrong.

Nadahar:
@Sami32 Please take a look at this now. You probably want more tests or have other things you want to change, but I consider myself more or less done. I only have some method renaming and of course the rebasing left.

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -305,9 +305,17 @@ HalveBitrate =
...
+H264LevelLimit =
+
+# The maximum supported H.265 level
+# Default:
+H265LevelLimit =
+
+# Whether to transcode H.265 video if it exceeds level 4.1
 # Default: true
-H264Level41Limited = 
+H265Level41Limited =
```
Nadahar: Hmm, it seems like some old stuff has sneaked in here - that shouldn't be here anymore..?

Sami32: Hehe, yes.
Trace of my original PR ;)

Sami32:
Maybe a common level class with some additional level values that could be used in future?
5.3, 5.4 and 6.3, 6.4

With 6.3 we are already in the 16K use case, so add an extra .4 could be an option though i feel like after that it will become 7.0.

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -1241,19 +1249,47 @@ public boolean isTranscodeAudioTo441() {
...
+			case H262:
+				return H262Level.typeOf(getString(codec.getLevelLimitKey(), null));
+			case H263:
+				return H263Level.typeOf(getString(codec.getLevelLimitKey(), null));
+			case H264:
+				return H264Level.typeOf(getString(codec.getLevelLimitKey(), null));
+			case H265:
+				return H265Level.typeOf(getString(codec.getLevelLimitKey(), null));
+			case MPEG4ASP:
+				return MPEG4VisualLevel.typeOf(getString(codec.getLevelLimitKey(), null));
```
Sami32: Do MPEG4SP not missing?

Nadahar: Yes, because I didn't find any levels for it :wink:

src/main/java/net/pms/media/MPEG4VisualLevel.java
```diff
@@ -0,0 +1,141 @@
...
+
+
+/**
+ * This represent the predefined MPEG-4 Visual levels.
+ *
+ * @author Nadahar
+ */
+public enum MPEG4VisualLevel implements VideoLevel {
+
+	/** Level 1 */
```
Nadahar: Add Level 0

Nadahar:
@Sami32 Now you can test it, I only have making and modifying tests left.

I'm very skeptical of the video level with remuxing with tsMuxeR code. The code is such a complete mess that I can't get an overview. I just guessed as best as I could... life is simply to short to fix `MEncoderVideo` and `FFMpegVideo`. I think rewriting them from scratch should be much faster than trying to make some sense of them. In either case, it's a huge project.

Sami32:
Your tsMuxeR seem correct to me, as PMS was forced to fix the level because they use a working Annex B code handling only that case IMO.

Probably, but refactoring the engines is not in my today plan ;-)

Sami32:
Work great :smile:

2 little things though:
- The logger for FFmpeg is buggy
>Not muxing the video stream with tsMuxeR via FFmpeg because the H.264 level (Level 5.1) is above the limit (Level 4.1) for this renderer

- The change in the log is too verbose and less readable IMO:
>Video Level: Level 4

 instead of "Video Level: 4"

I've to test again with MEncoder

Sami32:
Sorry, wrong paste:
>TRACE 2018-10-31 21:01:20.984 [HTTPv2 Request Worker 1] File "H.264" will not be streamed because the Level 5.1 level (Level 4.2) is above the limit ({}) for this renderer

Nadahar:
> The change in the log is too verbose and less readable IMO:
>
> `Video Level: Level 4`

I need some more info about where this is logged. Can you share the surrounding log? 

The reason it's this way is quite simple: Generalization. This code handles all video level now, and not all have a name like "Level 4". The names depend on the video codec. I have implemented two "variants" of `toString()`, one which returns the full name (like here) and one that returns the short/code form (which would be just "4", "4.1" etc.). The problem is that the short/code form will only reply some relatively cryptic codes like "LL", "H14" instead of "Low Level", "High 1440" for some of the other codecs.

It think it would work either way, but it's easy to change so I can do it if you can give me a hint about which code to change.

Nadahar:
Branch is rebased and requests have been implemented.

src/main/java/net/pms/encoders/MEncoderVideo.java
```diff
@@ -1378,7 +1398,12 @@ public ProcessWrapper launchTranscode(
 
 				encodeSettings = "-lavcopts " + aspectRatioLavcopts + vcodecString + acodec + abitrate +
 					":threads=" + configuration.getMencoderMaxThreads() +
-					":o=preset=superfast,crf=" + x264CRF + ",g=250,i_qfactor=0.71,qcomp=0.6,level=3.1,weightp=0,8x8dct=0,aq-strength=0,me_range=16";
+					":o=preset=superfast,crf=" + x264CRF + ",g=250,i_qfactor=0.71,qcomp=0.6,";
+				VideoLevel level = params.mediaRenderer.getVideoLevelLimit(VideoCodec.H264);
+				if (level != null) {
+					encodeSettings += "level=" + level.toString(false) + ",";
```
Sami32: In MEncoder, like in old FFmpeg versions, the level value need to be set as 2 digits only or it will fail.

Sami32: The level should be added this way:
... `:level=41:o=preset=superfast`

Nadahar: If that is the case, the "old" code was wrong as it used a dot: `level=3.1`. I only replicated that behavior. Why didn't it fail then?

Regarding the same situation in FFmpeg, I tried to find some information on the matter but didn't get very far. All I found is that some bugs that happen when using double digits won't happen when using the "dot version". Do you know how old versions of FFmpeg have to be to only accept the "non-dotted" version? Also, is level 4.0 written "4" or "40" in that case?

Nadahar: I did read this: https://lists.ffmpeg.org/pipermail/ffmpeg-user/2016-January/030084.html

Sami32: Yes `40`.
I just lloked at the MEncoder command line log and it clearly say that.
I saw that has been done before this PR.
No i don't know which FFmpeg version number did the trigger, but don't forget that some years ago their code was really similar.

Sami32: But it look that the FFmpeg level change has been recently added just before that issue report.

Sami32: From what i can see from FFmpeg code source the dotted values are only alias.

Nadahar: How do you explain the results in the linked issue if they are mere aliases?

Nadahar: About the MEncoder syntax, it looks like it's been like this for a long time: https://github.com/DigitalMediaServer/DigitalMediaServer/commit/3f9f9a820d1f350a2f31a435b4c1795a546b2030

Nadahar: Just to make things clear: It is no problem to change this, I just want to make sure what we do is correct.

Nadahar: I also wonder if we should use `-level:v` instead of `-level` for FFmpeg, and how far back that is supported.

Nadahar: The MEncoder parameter has been like this since the very beginning: https://github.com/DigitalMediaServer/DigitalMediaServer/commit/4570ae7760be974a1180de2119cb8929e31ba7be

Are you using a "SB" version when testing?

Sami32: Don't care for the video level `v` as we doesn't use it for audio and since it wasn't implemented in old FFmpeg versions.

Sami32: Yes

Nadahar:
Unable to verify if transcoding parameters are correct.

Sami32:
Ok ok i'll look at it today.
But it could have been handled in a separate PR as well.

Nadahar:
@Sami32 I've found another way.

PR #119, opened 2018-11-04
Web support enhancement

Sami32:
This change is [Reviewable](https://reviewable.io/reviews/digitalmediaserver/digitalmediaserver/119)


src/main/java/net/pms/configuration/WebRender.java
```diff
@@ -129,7 +130,7 @@ public boolean load(File f) {
 		configuration.addProperty(SUPPORTED, "f:jpg m:image/jpeg");
 		configuration.addProperty(SUPPORTED, "f:png m:image/png");
 		configuration.addProperty(SUPPORTED, "f:gif m:image/gif");
-		configuration.addProperty(SUPPORTED, "f:tiff m:image/tiff");
+//		configuration.addProperty(SUPPORTED, "f:tiff m:image/tiff");
```
Sami32: Should i need to comment it, or the below TIFF limitation is suffisant to avoid issue?

Nadahar: It doesn't matter, these aren't used anyway (no "supported" lines for images remember?).

Sami32: Memory flaw ;)

So shouldn't be better if i clean these supported lines?

Nadahar: The whole class (end the whole web interface stuff) needs a proper refactoring, so I don't care so much about the details either way.

Nadahar: But yeah, ideally, all the images could be removed.

Nadahar: If you change it, remember that I have rebased the branch so you must update your local first.

PR #148, opened 2024-02-16
Refactor Windows installer

Nadahar:
@Sami32 I guess it's time we merged this. I've updated it with the latest version of NSIS, working URLs and more, and done some "polishing". I think it should be good to go now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DigitalMediaServer/DigitalMediaServer/148)


pom.xml
```diff
@@ -1109,40 +1132,55 @@
...
-									<setupScript>${project.external-resources}/nsis/dms.nsi</setupScript>
+									<scriptFile>${project.external-resources}/nsis/${project.artifactId}.nsi</scriptFile>
+									<outputFile>${project.name.short}.exe</outputFile>
+									<compress>auto</compress>
+									<compression>lzma</compression>
+									<compressionIsFinal>true</compressionIsFinal>
+									<compressionDictSize>32</compressionDictSize>
+									<attachArtifact>false</attachArtifact>
 								</configuration>
-								<phase>package</phase>
```
Sami32, line 1128: I cannot remember well but make sure that the building time don't get negatively affected.

Nadahar: The phase is still set, it's just that I've set the default (correctly) in the plugin so there's no need to do it explicitly here:

https://github.com/DigitalMediaServer/nsis-maven-plugin/blob/34cd5087e57e9b712b80f96e92e469b5e4a821dd/src/main/java/org/digitalmediaserver/nsis/MakeMojo.java#L53

https://github.com/DigitalMediaServer/nsis-maven-plugin/blob/34cd5087e57e9b712b80f96e92e469b5e4a821dd/src/main/java/org/digitalmediaserver/nsis/GenerateHeaderfileMojo.java#L74

Nadahar (commented):
> Reviewed 471 of 471 files at r1, all commit messages.
> Dismissed @Sami32 from a discussion.
> *[Reviewable](https://reviewable.io/reviews/DigitalMediaServer/DigitalMediaServer/148)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nadahar)

Nadahar:
I found something that needs further investigation. The old installer copied the configuration files (DMS.conf, web.conf etc) to the profile folder, while it seems like this version only copies them to the installation folder. The installation folder isn't normally writable, so those are merely there as fallbacks. 

@Sami32 Do you remember why this change was done?

I think we might need to change something there - also it seems like it creates the profile folder in `ALLUSERSPROFILE` in the `.onInit` function. Doesn't that mean that the folder is created even if the installer is cancelled? If so, that's not good - no changes should be made if the installer is cancelled.

src/main/external-resources/nsis/setup.nsi
```diff
@@ -1,577 +1,790 @@
...
+	SetOutPath "$INSTDIR"
+	SetOverwrite on
+	WriteUninstaller "$INSTDIR\uninst.exe"
+
+	ReadEnvStr $R0 "ALLUSERSPROFILE"
+	SetOutPath "$R0\${PROJECT_NAME_CAMEL}"
+	CreateDirectory "$R0\${PROJECT_NAME_CAMEL}\data"
+	AccessControl::GrantOnFile "$R0\${PROJECT_NAME_CAMEL}" "(BU)" "GenericRead + GenericExecute + GenericWrite + Delete + FullAccess"
+	Pop $0
+SectionEnd
```
Nadahar, line 198: @Sami32 This is where you made these changes - here it used to copy the configuration files into the profile folder - and it used to create the folder itself.

Sami32: I can vaguely remember that these files need to be in the profile folder for Linux users to keep their settings.

Sami32: I can remember that these files was added in the uninstaller with their folder.

Sami32: It was a way to make the custom configuration kept among different Linux operating system.

Nadahar: NSIS isn't for Linux. The point is that they need to be in the profile folder to be editable for the user. We have "backup files" in the installation folder too, but they aren't meant to be edited - just to be there in case there's nothing in the profile folder. If the user tries to change a setting while using a configuration file in the installation folder, the change will fail.

So, we need to put them back. I can fix that, I just wondered if there was some particular "reason" that you could think of.

Nadahar:
I've discovered another problem too - the installer will create the "default installation folder" even if the installer is cancelled - and even if the user chooses a different folder. I think I understand why it happens, because there's a `SetOutPath "$INSTDIR"` in the `.onInit` function - and it seems to work fine when I comment it out. I'm just wondering what was the reason to put that there in the first place - it was probably to solve some other problem..?

src/main/external-resources/nsis/setup.nsi
Sami32: Sorry, yes, same reason for Windows.
As you said, without it users will not have their settings saved as soon as their session close.

Sami32: Well, better run it and test it before.
Because after 6/7 years my memory is not good...
The configurations files should be found in profile ALLUSERSPROFILE/DMS/data

I think that they will be there but the issue was that they couldn't if the folder was not set with the correct permissions (that probably all the care of the permissions enforced, and why it has been in .onInit phase.

Nadahar: No, I have tested, and the files are never copied there. That's no wonder since there's no code that tells NSIS to copy them there. All that is created is the folder. But, I've already fixed that so that's no problem - it was the reason I was after. But, it's probably lost to time.

Sami32: I though remember that it was the Java program part that was doing it so that why it was not added in this part.
But perhaps you have changed that part of the code that was duplicating the config files?
Anyway.

Nadahar: No, I don't think Java has ever done that - in the "old" installer the installer put the files there. It would of course be possible to do it from Java - but I don't know if there's any point in doing that instead.

Sami32 (approved):
OM.... ;p

Sami32 (approved):

Nadahar:
@Sami32 I've pushed a new commit, 06e5b48b75db929d3f598d0faa1297dad43b3d96, with my suggested changes. There are quite a few different things in there, so feel free to ask why.

Sami32 (dismissed):

